PR #21, opened 2016-04-29
Less Verbose make Commands (data, lint)

codyrioux:
By default pip will warn of requirements which are already installed, this fact coupled with the fact that `make data` depends on `requirements` (a good thing) generates a ton of uninteresting output for the task `make data`. I'm proposing pip be made silent as we aren't losing any transparency because make will still print `pip install -q -r requirements.txt` to the command line when executing `make data`.

The second change adds the virtualenv directories which will contain python code to the exclude list for flake8 so as to not generate linter warnings about code not written by the data scientists.

### DIFF
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -20,7 +20,7 @@ clean:
 	find . -name "*.pyc" -exec rm {} \;
 
 lint:
-	flake8 .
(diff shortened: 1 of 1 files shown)


isms:
@pjbull LGTM

pjbull:
Yeah, thanks @codyrioux, I like it!

isms:
Thanks!

Issue #18, opened 2016-04-29
Add explantion of using `.env` to docs


pjbull:
- E.g., deployment secrets, access keys, SQL connection strings

isms:
@pjbull just added explanation [here](http://drivendata.github.io/cookiecutter-data-science/#keep-secrets-and-configuration-out-of-version-control) via 873cfd6 - thoughts?

pjbull:
👏 👏 👏 

The one thought is that notebooks don't have `__file__` defined so we should try to make a code snippet that plays nicely with both.

<img width="1050" alt="screen shot 2016-04-29 at 3 13 04 pm" src="https://cloud.githubusercontent.com/assets/1799186/14928355/01ad03fe-0e1d-11e6-9ded-9fc2cb36c869.png">

isms:
Pull request added: https://github.com/drivendata/cookiecutter-data-science/pull/22
![](https://cloud.githubusercontent.com/assets/1284973/14932008/74c635ba-0e3e-11e6-9d9e-5219219fcdfc.png)

PR #22, opened 2016-04-29
Add util func to magically find .env file per #18

isms:
Add `find_dotenv` method that will try to find a .env file by
 1. guessing where to start using `__file__` or the working directory -- allowing this to work in non-file contexts such as IPython notebooks and the REPL, and then
 2. walking up the directory tree looking for the specified file -- called `.env` by default.

This is a bit like the ["filthy magic"](https://github.com/jpadilla/django-dotenv/blob/master/dotenv.py#L44-L46) employed by django-dotenv to serve the same purpose, and allows the user to write `load_dotenv(find_dotenv())` in many contexts.

I've submitted a [PR](https://github.com/theskumar/python-dotenv/pull/23) upstream to `python-dotenv`, but in the meantime it looks good for use in notebooks:
image

### DIFF
--- a/{{ cookiecutter.repo_name }}/src/data/make_dataset.py
+++ b/{{ cookiecutter.repo_name }}/src/data/make_dataset.py
@@ -4,6 +4,8 @@ import click
 import dotenv
 import logging
 
+from utils import find_dotenv
+
 
 @click.command()
 @click.argument('input_filepath', type=click.Path(exists=True))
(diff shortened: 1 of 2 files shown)


isms:
@pjbull Disregard! @theskumar worked lightning fast and merged my [PR](https://github.com/theskumar/python-dotenv/pull/23) into a new release. Will bump requirement up to `python-dotenv==0.5.0`.

Closing.

PR #29, opened 2016-05-24
Removed --recursive flag from s3 sync in the Makefile

mrbell:


pjbull:
Thanks @mrbell! Let us know if you spot anything else amiss.

PR #35, opened 2016-07-06
Update .gitignore file to ignore the data folder

liudonghs:
I guess the correct way of ignoring the data folder is: `data/` rather than `./data/`


drivendata:
Yep, thanks!

PR #44, opened 2016-08-06
Make `make lint` work for new projects.

hwartig:
With this patch `make lint` wouldn't report errors in newly created projects.


pjbull:
Good catch, thanks!

PR #45, opened 2016-08-14
Update doc to reflect AWS CLI rather than s3cmd

adamkgoldfarb:
Currently the [documentation page](https://drivendata.github.io/cookiecutter-data-science/#cookiecutter-data-science) lists s3cmd as being used rather than AWS CLI, which is what is used in the Makefile. Changing the documentation would likely save some folks from combing through s3cmd so they could instead head directly to the docs for AWS CLI.


pjbull:
Awesome, thanks @adamkgoldfarb! Would you mind making "AWS CLI" a link to those docs?

adamkgoldfarb:
Not a problem, all set! I referred the S3-specific docs within the broader AWS CLI docs. Thanks @pjbull for the good work!

pjbull:
💃

Issue #14, opened 2016-04-26
Add description of virtualenv tools to the documentation


jbrambleDC:
I believe Conda can handle what virtualenv + pip is doing in a better way? Should we consider adding this functionality?

pjbull:
Unfortunately, the conda environments are not cross-platform, so we generally prefer pip + virtualenv.  From the [conda docs](http://conda.pydata.org/docs/using/envs.html#create-a-separate-environment):

> NOTE: These explicit spec files are not usually cross platform, and therefore have a comment at the top such as # platform: osx-64 showing the platform where they were created. This platform is the one where this spec file is known to work. On other platforms, the packages specified might not be available or dependencies might be missing for some of the key packages already in the spec.

This template could end up conda/virtualenv agnostic if this issue gets fixed:
https://github.com/conda/conda/issues/1277

Closing for now. 😦

isms:
@jbrambleDC Hey Jordan, thanks for contributing! Just for the record, I almost always use conda environments which are admittedly kind of fiddly to get set up for reasons described pretty well in conda/conda#292 ... that being said it's not too bad.

Here's a hacky workaround I sometimes use:

```bash
# install all binary packages through conda, fail on ones that aren't available
$ cat requirements.txt | while read package; do conda install $package; done
[... blah blah blah, lots of installation]
# now install the rest via the conda env's pip
$ which pip
/home/isaac/anaconda2/envs/test/bin/pip
$ pip install -r requirements.txt 
Requirement already satisfied (use --upgrade to upgrade): click in /home/isaac/anaconda2/envs/test/lib/python2.7/site-packages (from -r requirements.txt (line 1))
[... conda ones are already installed, others get installed now]
```

pjbull:
I think that we may want to give the option to use a `conda env` over `virtualenv`--for example, i just had a project with complex dependencies (e.g., GDAL) and the conda env worked well for that. I also think their env stuff is getting better overall--for example, they now have a path to fall back to `pip`.

Reopening to track adding as an option.

isms:
@jamestwebber I use the template daily and use conda almost exclusively so I don't think there's a conflict.

If we're giving the impression that there is a conflict it'd be helpful to know how in order to fix that.

jamestwebber:
Does it also create a virtualenv inside of a conda environment?

[...]

PR #50, opened 2016-10-05
Makefile help, specify python interpreter, test environment

niloch:
This pull request adds  several enhancements:

* [Default rule](https://gist.github.com/klmr/575726c7e05d8780505a) in the Makefile to show the other commands with their descriptions from the comments above.
* Additional variable to the template for setting a python interpreter (python vs python3)
* 2 new Makefile commands
     - `create_environment` new virtualenv with the correct interpreter using `virtualenvwrapper`(optional)
     - `test_environment` tests the python environment before installing requirements

### DIFF
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -5,13 +5,15 @@
 #################################################################################
 
 BUCKET = {{ cookiecutter.s3_bucket }}
+PROJECT_NAME = {{ cookiecutter.repo_name }}
+PYTHON_INTERPRETER = {{ cookiecutter.python_interpreter }}
 
 #################################################################################
 # COMMANDS                                                                      #
 #################################################################################
 
 ## Install Python Dependencies
-requirements:
+requirements: test_environment
 	pip install -q -r requirements.txt
 
 ## Make Dataset
@@ -33,6 +35,15 @@ sync_data_to_s3:
 ## Download Data from S3
 sync_data_from_s3:
 	aws s3 sync s3://$(BUCKET)/data/ data/
+
+## Set up python interpreter environment
+setup_environment:
+	source /usr/local/bin/virtualenvwrapper.sh; mkvirtualenv $(PROJECT_NAME) --python=$(PYTHON_INTERPRETER)
+	@echo ">>> Activate virtualenv with:\nworkon $(PROJECT_NAME)"
+
+## Test python environment is setup correctly
+test_environment:
+	python test_environment.py
 
 #################################################################################
 # PROJECT RULES                                                                 #
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -10,24 +10,93 @@ BUCKET = {{ cookiecutter.s3_bucket }}
 # COMMANDS                                                                      #
 #################################################################################
 
+## Install Python Dependencies
 requirements:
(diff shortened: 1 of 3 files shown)


{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -5,29 +5,109 @@
...
 sync_data_to_s3:
 	aws s3 sync data/ s3://$(BUCKET)/data/
 
+## Download Data from S3
 sync_data_from_s3:
 	aws s3 sync s3://$(BUCKET)/data/ data/
 
+## Set up python interpreter environment
+create_environment:
+	source /usr/local/bin/virtualenvwrapper.sh; mkvirtualenv $(PROJECT_NAME) --python=$(PYTHON_INTERPRETER)
```
pjbull: Will this path exist on OSX?

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -5,29 +5,109 @@
 #################################################################################
 
 BUCKET = {{ cookiecutter.s3_bucket }}
+PROJECT_NAME = {{ cookiecutter.repo_name }}
+PYTHON_INTERPRETER = {{ cookiecutter.python_interpreter }}
```
pjbull: Can we use this variable for the Make commands that call Python?

pjbull:
That self-documenting Make magic is ✨, thanks for submitting!

I'm pro on defaulting to `virtualenv`, but someday we'll want to support conda env as well. Reopening #14 to track that.

{{ cookiecutter.repo_name }}/Makefile
pjbull: Yep, agreed. Let's output the required profile lines to the console so users see them. I don't think there's a really elegant way to do this otherwise.

In general, for users who are opinionated about their environment tool, it should be easy to change. For those who care less, using the default tool in CCDS should be as painless as possible. Right now, I think that virtualenvwrapper is the least painful (primarily in that it does not require the Anaconda distribution of Python).

{{ cookiecutter.repo_name }}/Makefile
pjbull: Yeah, that seems right but there's an added wrinkle for projects that don't use virtualenv:
we ask for the Python interpreter during setup without suggesting (or requiring) that that interpreter be used to setup a virtualenv. In that case, you might expect that setting the Python interpreter be used for any commands where Python gets called.

So, I think there are two options that would work:
 1) `[OPTIONAL] Choose Python interpreter for virtualenv`
 2) Use the interpreter passed in setup for Make commands. I haven't tested this, but I assume that if I do `--python=python3` when I create my virtualenv, calling `python3` at the command line will still get the right one when I have that env active.

I'm leaning towards 2 since it covers both use cases. Thoughts?

niloch:
Added `pip install -q virtualenvwrapper` to the `create_environment` command.  I tried to use bash conditionals to detect whether it was already installed or not, and if so, to suppress the message about updating your shell startup file, but I couldn't get it to work.

pjbull:
Thanks @niloch!

I added support for detecting Anaconda and not creating virtualenvs and also had to make a couple tweaks for this to run for me on an Ubuntu VM. If you get a chance would you mind checking if the latest on master still works for your case?
https://github.com/drivendata/cookiecutter-data-science/commit/1caa67be438fec5a177e158d56bedade0854faff

niloch:
Everything looks great!  Thanks for merging.

PR #57, opened 2016-11-10
Changed findstr to findstring

verginer:
On Mac with GNU Make 3.81 I had to change `findstr` to `findstring` to get the makefile to correctly read the python version number.


pjbull:
@verginer Thanks!

PR #63, opened 2017-02-22
Check if anaconda is installed with project chosen python version (and not always python2) 

keldLundgaard:


pjbull:
This looks fine to me, but I am curious about the machine setup when you hit an issue.

We use this check just to see if Anaconda is installed (and, what we actually check is if the `python` alias executes an Anaconda version of python). It shouldn't matter if the `python` alias runs 2 or 3. In fact, this call will run Python 3 if that is your default environment on your Anaconda install.

So, summary, I think this will work fine, but I'm not sure when it would be necessary.

keldLundgaard:
Cool. 

It made an error on both mine and my coworkers macs with the newest anaconda with python3 installed.

pjbull:
Can you share the error?

keldLundgaard:
Sorry, error was not the right word. It would not find anaconda when using python instead of python3, so it would think that it was not installed.

keldLundgaard:
Are you accepting this or do you want to explore it further?

pjbull:
There is one edge case I'm concerned about:
 - Install Anaconda (Python 2.7)
 - Run cookiecutter-data-science with interpreter as `python3`

I _believe_ that this fix will fail in the scenario since `python3` won't be on the `PATH` anywhere. However, it's a valid use case to want to bootstrap a Python3 environment even if your system python and Anaconda are both Python2.

It might be a better fix just to test if `which conda` returns anything.

Off the top of my head (untested), this may work:
```
ifeq(,$(shell which conda))
IS_ANACONDA=False
else
IS_ANACONDA=True
endif
```

keldLundgaard:
Yes that seems like a better solution. I'll close this and make a PR based on that.

Issue #56, opened 2016-10-25
Initialize Git repository on creation


mattayes:
Is it possible for CookieCutter to run a command like `git init` after repository creation? It's not a make-or-break, just a nice convenience.

isms:
This is a cool idea. Two potential concerns here:
1. Git-centricity. Even though this project uses git+Github as the VCS, we actually don't make assumptions about what version control the *data science project* will be using.
2. Platform compatibility. Even if we're not directly calling out using `subprocess` or the link (and assuming a binary called `git` is on the system path), then even using a Python git wrapper I fear we are assuming `libgit` is installed.

I'm inclined to give this a discussion label for a while and see if people feel strongly.

mattayes:
That all makes total sense. I think you're right, it's wrong to assume everyone is going to use Git.

It might be nice to give users a choice of which VCS they want to use (or do nothing if no input is provied) and then roll from there. We still run into a platform compatibility issue, though it's a little more manageable; you shouldn't be picking a VCS you don't use/have installed (if you try to, though, things will still get messed up).

What do you think?

ConstantinoSchillebeeckx:
For those interested, I'm doing this in my fork of the cookiecutter with the [following bash script](https://github.com/ConstantinoSchillebeeckx/cookiecutter-data-science/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/src/setup/setup_git_repo.sh)

I then added a command to the Makefile:
```
## Push to GitHub repo
push: 
	git add -A
	git commit -m 'Make Git update'
	git push
```

mattayes:
Nice, I like it @ConstantinoSchillebeeckx!

@isms What do you think about a plug-and-play way to initialize a version control system?

isms:
@mattayes Cool idea - what do you think an implementation might look like?

pjbull:
Also, just a note here, I've been holding off on optional features while I watch:
https://github.com/audreyr/cookiecutter/pull/848

Unfortunately, that issue is currently stalled because the cookiecutter project needs sponsorship. 

For things like VCS where we want options and may have fields that are conditional on your choice, it's pretty tough to manage with the current cookiecutter workflow, which doesn't really do branching. I was hoping that the PR referenced above would get pulled in for these kinds of features, but that's not super likely.

The one thing we could look at is doing is conditional files using this kind of syntax:
https://github.com/audreyr/cookiecutter/issues/723

[...]

PR #64, opened 2017-02-24
Test for conda.

keldLundgaard:


keldLundgaard:
An a semi-related note: Why the fallback on virtualenv. Couldn't it just declare that you should install anaconda or miniconda?

pjbull:
LGTM, merging.

Not every not every environment/organization can install Anaconda, so we fall back to virtualenv.

Ultimately, it would be nice to just pick one when you run `cookiecutter`, but that's sort of dependent on the issues with cookiecutter that are mentioned here:
https://github.com/drivendata/cookiecutter-data-science/issues/56#issuecomment-277765734

Issue #10, opened 2016-04-23
Make default README better


drivendata:

marskwd:
You may want to update [tox.testrun.org](https://tox.testrun.org/) link to [tox.readthedocs.io](https://tox.readthedocs.io/en/latest/) The security certificate expired, and the browser warnings threw me off. Not sure whether this is just an issue with my browser or if everyone knows what tox is and haven't had to look this up.

PR #75, opened 2017-06-14
Add directory structure to README, improves #10

lorey:
The file structure is an essential part of the provided benefit and should be in the readme file.


isms:
Merged - thanks @lorey!

PR #78, opened 2017-06-28
Use latest stable python3 in conda create

KPLauritzen:
When running `make create_environment` with `PYTHON_INTERPRETER = python3` and `HAS_CONDA=True`, we should use the latest stable python3, instead of python3.5


pjbull:
Thanks @KPLauritzen, good call. On to 3.6! 💯

PR #80, opened 2017-07-03
adds .editorconfig

verginer:
Adding an [.editorconfig](http://editorconfig.org/) file as is also included in [audreyr/cookiecutter-pypackage/](https://github.com/audreyr/cookiecutter-pypackage/) I think might be a very simple and non intrusive way to introduce more consistency in the code.


verginer:
Sorry wrong branch will open a pull request only with editorconfig

Issue #72, opened 2017-05-27
Slightly adjust commands in Makefile


Midnighter:
Just started using your cookiecutter for the first time. Thank you for the effort, it seems very valuable to me!

I had some comments and I'm happy to create pull requests if those are desired changes. I'm talking about targets in the Makefile here.

`requirements`:

pip now also [allows a constraints file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) which seems more relevant to me in order to pin or require certain versions of dependencies.

`clean`:

Find can delete directly: `find . -iname "*.pyc" -delete` seems pretty clear to me. For Python 3 it could be useful to add `find . -iname "__pycache__" -exec rm -rf {} +`. The plus at the end, rather than the `\;` will pipe all found instances to the `rm` tool in one go rather than executing `rm` for each one individually.

`lint`:

Typically I only want to run flake8 on source code. So rather than excluding a bunch of directories why not specifically call flake8 on the `src` directory only?

pjbull:
Thanks for the report @Midnighter!

The changes to `clean` and `lint` seem good to me unless @isms had a thought as to why we would `lint` other directories. Also I'm a little wary of a `rm -rf` hidden in a `Make` command, so maybe we do both and remove the empty `__pycache__` with `rmdir` instead.

WRT to `requirements` my reading of the doc is that `requirements.txt` is still where requirements will live by convention (including pinning certain versions). It seems that the constraints file is if you want to enforce certain versions across project, for example, you can include one that will automate the checking that the project meets the requirements. However, it doesn't install the requirements. I'd leave as is.

isms:
Thanks @Midnighter. Chiming in with some thoughts:

`requirements`: this is a nice point but I vote we revisit when `constraints.txt` is more widespread.

`clean`: no preference since the `-delete` flag is sort of `rm -rf`ish and it shouldn't be a big deal with the pattern match.

`lint` concur, let's just lint `src/`

PR #84, opened 2017-07-21
feat: upgrade essentials before requirement

Midnighter:
This is one of a series of PRs that relate to #72.

Just to be sure I think it's nice to have the essential Python installation tools up-to-date all the time.


isms:
Good call.

PR #91, opened 2017-12-13
Include gitkeep's

ayemos:
`.gitignore` file in generated cookie excludes whole `/data/` which make all `.gitkeep`s under `/data/external/, /data/interim/ etc.` being ignored.

This PR will let `.gitignore` ignore whole data under `/data/` directory except `.gitkeep`s under four specific data directories `external, interim, raw, processed`.

please review 🙏

### DIFF
--- a/{{ cookiecutter.repo_name }}/.gitignore
+++ b/{{ cookiecutter.repo_name }}/.gitignore
@@ -70,4 +70,9 @@ target/
 .ipynb_checkpoints/
 
 # exclude data from source control by default
-/data/
+/data/**
+!/data/external/
+!/data/interim/
+!/data/raw/
+!/data/processed/
+!/data/*/.gitkeep


lunardog:
🤔 
Even if `data/raw` has no `.gitkeep`it will *not* be ignored with this `.gitignore`. 
Is this the expected behavior?

ayemos:
the directories won't be ignored but never checked in unless there's any files under them.
I think there's no simple way to do "conditional ignore".

ref: https://gist.github.com/ayemos/226d4e4def0cb61086d4b94ca2377ea6

pjbull:
Open to other ideas here, but wanted to share what I believe the _intended_ behavior was in the existing implementation.

 - One of the tenets of the project is that for most projects data should not live in the source repository by default.
 - Despite not including the data in the repository, we want to make recommendations about how and where to keep your on the file system with respect to the project.
 - With that in mind, the `.gitkeep`s are intended to create the `data` folder and its subdirectories when a new project is initiated with the cookie cutter.
 - After initiation, the contents of the data folder should be excluded from source control by default.

TLDR; original design was `data/` and its directories in the _template_ repo, but not necessarily any instantiated project's source control.

Again, open to thoughts/opinions/recommendations, but IIRC this was the original intent.

isms:
Closing in light of @pjbull's comment but @ayemos happy to reopen if you have a different take on that.

ayemos:
> original design was data/ and its directories in the template repo, but not necessarily any instantiated project's source control.

Understood, thank you for the comment :)

PR #103, opened 2018-03-14
Offer accurate, open-ended licensing suggestions.

### LINKED ISSUE #102: Present licensing options honestly.
Hello!

I find that the way licensing options are presented is not honest: https://github.com/drivendata/cookiecutter-data-science/blob/dc146594089677ecc80f85f122ff3ab788eeaf4d/cookiecutter.json#L6

It is not open-ended, so it is basically stating that MIT and BSD (which one?) are the only "open source" licenses out there. There are many more FLOSS licenses to choose from: https://opensource.org/licenses/

I'm happy to prepare a PR to reword all this.

Thank you,
Marianne

mkcor:
Hi @pjbull,

This change alone would fix #102! Not that we couldn't take it even further...

Thanks for sharing your thoughts.

Best,
Marianne

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -3,7 +3,7 @@
     "repo_name": "{{ cookiecutter.project_name.lower().replace(' ', '_') }}",
     "author_name": "Your name (or your organization/company/team)",
     "description": "A short description of the project.",
-    "open_source_license": ["MIT", "BSD-3-Clause", "Other"],
+    "open_source_license": ["MIT", "BSD-3-Clause", "No license file"],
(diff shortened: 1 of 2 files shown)


pjbull:
Thanks!

Could you also add the BSD 3-clause license text in the `LICENSE` file? (currently selecting that option does nothing)
https://github.com/drivendata/cookiecutter-data-science/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/LICENSE#L1

Also, let's say "No license file" as the third option, just to be explicit.

mkcor:
@pjbull Right, I thought about the `{{ cookiecutter.repo_name }}/LICENSE` afterwards! Done.

pjbull:
Thanks @mkcor!

Issue #101, opened 2018-02-28
Turn src into a python package rather than collection of scripts


pjbull:
More and more, I'm starting to feel like it will be more valuable to everyone to have `src` be a proper package rather than a collection of scripts.

Advantages:
 - Can consistently import and run package (rather than jumping through hoops based on if the file is run as a script or a module).
 - Will fix the hack of mucking with the `sys.path` to import in notebooks (though will probably still need autoreload to bust in-memory package caching ) 
 - Will better prepare the code to be extracted, generalized, and distributed if that becomes useful.

Disadvantages:
 - Increased complexity + boilerplate
 - Pushes us farther down the path of a Python-specific template (which I think we've accepted anyway)
 - Can be more confusing for beginners used to running `python XXXXX.py` from the commandline.

I think that we can automate away the more painful parts by templating `setup.py` and using `pip install -e src` during the requirements.

Opening this issue to discussion.

isms:
Closed by #104

PR #104, opened 2018-03-23
Create and install src as Python package

pjbull:
@isms I think this is the minimum bar that addresses #101.

If it looks good, I'll add some tweaks to the docs as well to update the notebook imports.


pjbull:
@isms Ping

PR #106, opened 2018-03-30
Added _.DS_Store_ OSX directory file extension

randallrs:
These files are created when new folders are created within MacOSX directories.


pjbull:
Thanks!

PR #110, opened 2018-04-14
Add S3 prefix configuration

### LINKED ISSUE #109: Add to docs that S3 can be a path (rather than just a bucket name)
I think it's a bit of an odd restriction to require each project to have their own bucket in S3. Users should also be able to optionally specify a prefix for their files in S3 besides `/data/`. 

This way users could have `s3://my-data-science-bucket/project1/data/`

Raab70:
Closes #109

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -5,7 +5,7 @@
     "description": "A short description of the project.",
     "open_source_license": ["MIT", "BSD-3-Clause", "No license file"],
     "s3_bucket": "[OPTIONAL] your-bucket-for-syncing-data (do not include 's3://')",
-    "s3_prefix": "[OPTIONAL] prefix of your data in S3",
+    "s3_prefix": "[OPTIONAL] prefix of your data in S3 (data will be synced to s3://your-bucket-for-syncing-data/your-prefix/",
     "aws_profile": "default",
     "python_interpreter": ["python", "python3"]
 }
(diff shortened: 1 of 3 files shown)


cookiecutter.json
```diff
@@ -5,6 +5,7 @@
     "description": "A short description of the project.",
     "open_source_license": ["MIT", "BSD-3-Clause", "No license file"],
     "s3_bucket": "[OPTIONAL] your-bucket-for-syncing-data (do not include 's3://')",
+    "s3_prefix": "[OPTIONAL] prefix of your data in S3",
```
isms: This is not descriptive enough - can you expand so that people know what a prefix is and what a reasonable choice might me?

isms: Please add a default prefix of `data/`.

Raab70: I wanted to add a default of `data/` but didn't see an option to. Would this just be in the Makefile if there is no prefix set it to `data/`

Raab70:
I added some minor clarification but I hestitate to put much more in that single line. I'd be happy to add a section to the docs with some more information about how the sync will look with some examples. Maybe under the getting started heading?

pjbull:
@isms @Raab70 I'm don't want to add a new parameter to the default flow. Seems like someone could easily put `bucket/folder` in the bucket param and that will work. It would be reasonable to add that to the docs or even update to `s3_path` rather than `s3_bucket`.

Raab70:
Good call @pjbull, I'm not sure how I missed that. I would suggest updating the docs as this is definitely not clear but makes sense now.

PR #112, opened 2018-04-15
Use pathlib instead of os.path where applicable

### LINKED ISSUE #111: Replace all uses os.path with pathlib
All hail the [pathlib](https://docs.python.org/3/library/pathlib.html) :sparkles::sparkles::sparkles:

isms:
Closes #111

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -1,93 +1,86 @@
 import os
+import pytest
 import shutil
+from pathlib import Path
 
 from cookiecutter import main
-import pytest
 
-CCDS_ROOT = os.path.abspath(
-                os.path.join(
-                    __file__,
-                    os.pardir,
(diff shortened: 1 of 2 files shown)


pjbull:
@isms I think we need pathlib in `requirements.txt` since it's not in Python 2 😢 
http://python-future.org/imports.html#included-full-backports

isms:
@pjbull Good point... two questions though:

- Do we need to support a Python2 example stub and `test_[stuff].py`?
- What if we're explicit with a `#!/usr/bin/env python3` at the top of those files?

Would either of those fix the issue?

(I'd also be interested to hear from the community how many people are still starting new projects with 2.x.)

isms:
(Just pushed a conditional requirement @ 74867b0 to address the issue in the meantime!)

PR #113, opened 2018-04-19
Make python3 the default interpreter

isms:


pjbull:
I'm all for this, but in the long term I'm thinking we should ask for an X.X version number instead.

PR #128, opened 2018-07-07
Using change-tracking to avoid re-running pipeline

jkabrg:
One of the main applications of `make` is to stop people from running a pipeline over and over again.
Consider the pipeline A -> B -> C -> D. If step C changes, only steps C and D need to be re-run, not steps A and B. Currently, this makefile will rerun steps A and B. This pull request starts to change that.


jkabrg:
Did a `rebase` and `git push --force` on the above

PR #129, opened 2018-07-08
Change version in setup.py to 0.1.0

JohnPaton:
In `setup.py` the version is currently `0.0.1`. For [semantic versioning](https://semver.org/) the version numbers should correspond to `MAJOR.MINOR.PATCH`, with `PATCH` only being incremented for bugfixes. Since the initial development is not a bugfix but the first minor version, I think it would help encourage best practices in versioning to change the initial version in the template to `0.1.0`.

Thoughts?


pjbull:
Yep, that looks right according to the SemVer FAQs. Thanks for raising it!

PR #131, opened 2018-07-25
Expand Setup Tests

dmitrypolo:
Part of our Data Days for Good initiative was to work on this repo. One of the requested additions was fleshing out the setup tests a little bit more to test for what would happen if user input was introduced. This was mentioned by Isaac. I have separated out the fixture and put it in it's own file, which is discoverable by `pytest` and mentioned in the docs. I have also increased the scope of creating the temp directory to be a class scope so it doesn't have to create it and tear it down after every function call. Lastly I also test for certain conditions given user input.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ docs/site/
 # test cache
 .cache/*
 tests/__pycache__/*
+*.pytest_cache/
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,35 @@
+import pytest
+import shutil
+from pathlib import Path
+from cookiecutter import main
+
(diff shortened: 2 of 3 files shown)


isms:
@dmitrypolo Will review today, thanks!

dmitrypolo:
I had the name backwards, for the assertion, can you please try again? If it fails on the temp file creation then I might have to look into it more as it might be a difference between `Darwin` and `Linux`

dmitrypolo:
seems like it is an issue between Linux and Mac, just hopped onto a Linux box and verified the same errors are happening. I will investigate and submit a fix when ready. Thanks for pointing this out. @isms

isms:
Merged, thanks @dmitrypolo!

PR #135, opened 2018-08-01
New dataset

johnkarlen:
switching to the transfusion dataset made a few features for years and quarters for the dataset


{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -29,6 +29,14 @@ requirements: test_environment
 data: requirements
 	$(PYTHON_INTERPRETER) src/data/make_dataset.py
 
+## Train Model
```
isms: These should only be included if the option is selected. Please use a template conditional.

cookiecutter.json
```diff
@@ -6,5 +6,6 @@
     "open_source_license": ["MIT", "BSD-3-Clause", "No license file"],
     "s3_bucket": "[OPTIONAL] your-bucket-for-syncing-data (do not include 's3://')",
     "aws_profile": "default",
-    "python_interpreter": ["python3", "python"]
+    "python_interpreter": ["python3", "python"],
+    "include_starter_proj": ["N", "Y"]
```
isms: Can we make these lowercase and also use `cookiecutter.include_starter_proj|lower` when we're doing checks so it doesn't matter whether the use inputs `y` or `Y`?

{{ cookiecutter.repo_name }}/setup.py
```diff
@@ -4,7 +4,8 @@
     name='src',
     packages=find_packages(),
     version='0.1.0',
+    install_requires=['{% if cookiecutter.include_starter_proj == "Y" %}scikit-learn[alldeps]{% endif %}'],
```
isms: I think this is overly complex - is it enough to just put `scikit-learn` in the requirements.txt?

Also there's a bug here in that the `if` is inside the quotes so you'll end up with `install_requires=['']` if the conditional isn't met.

{{ cookiecutter.repo_name }}/src/data/make_dataset.py
```diff
@@ -1,30 +1,47 @@
-# -*- coding: utf-8 -*-
-import click
-import logging
+import os
+import io
+import requests
+import pandas as pd
 from pathlib import Path
-from dotenv import find_dotenv, load_dotenv
```
isms: We'd prefer not to remove this (and the usage below) as it shows how to use a `.env` safely for credentials.

{{ cookiecutter.repo_name }}/src/data/make_dataset.py
```diff
@@ -1,30 +1,47 @@
...
-    log_fmt = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
-    logging.basicConfig(level=logging.INFO, format=log_fmt)
 
-    # not used in this stub but often useful for finding various files
-    project_dir = Path(__file__).resolve().parents[2]
+def main():
+    """ Retrieves data and runs processing scripts to turn raw data from (../raw) into
+        cleaned data ready to be analyzed (saved in ../processed).
+    """
+    s=requests.get(DATA_LINK).content
```
isms: The downloading should happen with a `Makefile` rule (using `curl` or `wget`), using requests to `StringIO` works for tiny files but is a bad practice to show to learners.

{{ cookiecutter.repo_name }}/src/models/predict_model.py
```diff
@@ -0,0 +1,32 @@
...
+    """ retrieve the model and predict labels. Show prediction and performance
+    """
+    deserialized_model = retrieve_model()
+    X_test = pd.read_csv(os.path.join(ROOT, 
+        'data/processed/transfusion_x_test.csv'))
+    y_pred = deserialized_model.predict(X_test)
+
+    y_test = pd.read_csv(os.path.join(ROOT,
+        'data/processed/transfusion_y_test.csv'), header=None)
+    print(f'The model returned these predictions:\n{y_pred}')
```
isms: Let's use proper logging statements instead of `print()`

{{ cookiecutter.repo_name }}/src/models/train_model.py
```diff
@@ -0,0 +1,59 @@
+import os
+import pickle
+import pandas as pd
+from pathlib import Path
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.model_selection import train_test_split
+
+ROOT = Path(__file__).resolve().parents[2]
+
+if 'python3' == 'python3':
```
isms: `if 'python3' == 'python3'`? Did this get templated before being committed?

{{ cookiecutter.repo_name }}/src/models/train_model.py
```diff
@@ -0,0 +1,59 @@
...
+
+def main():
+    """ Trains the model on the retrieved data write it back to file
+    """
+    x_train, x_test, y_train, y_test = fetch_processed('data/processed/transfusion_data.csv')
+    # Train the model 
+    model = fit_model(x_train, y_train)
+
+    # Paths for storage
+    model_out_dir = os.path.join(ROOT, 'models/transfusion.model')
```
isms: Since `ROOT` is a `pathlib.Path` object you can just do this:

```python
model_out_dir = ROOT / 'models/transfusion.model'
x_test_path = ROOT / 'data/processed/transfusion_x_test.csv'
y_test_path = ROOT / 'data/processed/transfusion_y_test.csv'

# Store model and test set for prediction
with model_out_dir.open('wb') as fout:
    pickle.dump(model, fout, PROTOCOL)
...
```

{{ cookiecutter.repo_name }}/src/models/predict_model.py
```diff
@@ -0,0 +1,32 @@
...
+import pickle
+import pandas as pd
+from sklearn.metrics import roc_auc_score
+from train_model import ROOT
+
+
+def retrieve_model():
+    """retrieve the pickled model object
+    """
+    pickled_model = os.path.join(ROOT, 'models/transfusion.model')
```
isms: Please use the `pathlib.Path` idioms for joining (see other comment below)

isms:
@johnkarlen Why closed?

PR #148, opened 2018-10-25
Add `module_name` option to replace `src`

hackalog:
Allow the user to change the name of the default module (currently `src`).

I've been running with this change for a while.


yahiakala:
This seems like a no-brainer to me. At the very least, it should not be named something so generic like src. Is there anything else that needs to be done to make this happen?

pjbull:
Yes, will be implemented as part of #162, thanks for the PR!

PR #149, opened 2018-10-31
Update link to snakemake

andrewsanchez:
Link to snakemake's readthedocs.io instead of bitbucket


isms:
Thanks @andrewsanchez!

PR #150, opened 2018-11-10
Update Makefile

unai-zulaika:
Work with pip from $PYTHON_INTERPRETER as recommended in 
https://packaging.python.org/guides/installing-using-pip-and-virtualenv/


pjbull:
Thanks @unai-zulaika!

PR #151, opened 2018-11-13
fixed gitignore typo

luke-soteris:
Unless there is some odd comma syntax I don't know about, it seems there is a typo in the .gitignore file.


pjbull:
@luke-soteris Thanks!

Issue #156, opened 2018-12-16
integration with mlflows and snakemake


chanansh:
I think that logging via mlfows and make via snakemake can enhance this project

isms:
Copying a reply from #158:

Our approach to integrations is generally to leave that up to the user.

Some questions that might be helpful to hear more about:

* If somebody intending to use DVC can just dvc init in their project then what would an integration look like?
* And if they don't use or know about DVC is the purpose of the integration to push them to use it or educate about its merits? How might this affect the user experience?

Issue #140, opened 2018-08-30
Rename `src` directory to `project_name`


oczkoisse:
First off, thanks for the wonderful template! I think this should work very well for my future projects. However, I could appreciate your thoughts on some aspects of the template:

- Filenames like `make_features.py` and `make_dataset.py` in the `src` package suggest that those are scripts, not modules. Wouldn't it be better separation to let the reusable code live in modules inside `src` package while the scripts reside in a `scripts` directory in project root? This way both the scripts and the notebooks could use the modules inside the `src` package. Could you please elaborate the rationale behind the current template layout, which seems to have just scripts in the `src` package?

- If I make my own project-specific packages, do they go under `src`? If I have a custom package that models, say, the dataset structure, does it make more sense to put it under `src/data` instead?

One minor nitpick: the name `src` could have been renamed to the project name itself. It would be clearer, I think, and also integrate well with a popular project structure suggested [here](https://docs.python-guide.org/writing/structure/), although I do understand that it may be a matter of personal preference.

TobiasSkovgaardJepsen:
As someone who recently discovered this template:

I had the same initial thought, but I think (and someone please correct me if I am wrong) that one should view the `src` subfolders (data, models, features) as separate apps. In that case, it makes sense that the `make_dataset.py` and `make_features.py` are stored in the folder specific to the app as a "main" file. This is similar to how the web framework Django does it for management commands: https://simpleisbetterthancomplex.com/tutorial/2018/08/27/how-to-create-custom-django-management-commands.html

> One minor nitpick: the name src could have been renamed to the project name itself. It would be clearer, I think, and also integrate well with a popular project structure suggested here, although I do understand that it may be a matter of personal preference.

As you suggest, this is probably a matter of personal preference. For instance, I would usually call such a folder `python` because I usually also have a significant amount of SQL and bash scripts in `sql` and `bash` folders, respectively.

pjbull:
(1) Agreed that `src` -> `{{project name}}` makes sense. We'd accept that PR.
(2) At times it can be useful to distinguish scripts from packages. In our experience, we don't find that we need this distinction in most of the ML/DS projects we do. Often our code that ends up in a package also has a CLI that you can enter like a script. If projects get built up to the point where they have scripts and packages, users can create a new folder. Just not sure there is a strong case for this to be built in as boilerplate for the default. I think the simplicity of one folder for source code is probably best for now.

bangar1991:
[...]

Issue #120, opened 2018-06-05
Google Cloud Storage support


jtascensao:
Hi guys! I'm wondering if you guys have any plans to add Google Cloud Storage support, in a similar way to what you already do with S3 and AWS.

That would be extremely helpful!

dkapitan:
@jtascensao would be interested in this as well. We work on Google Cloud platform, where every project has a separate:
- Cloud Storage bucket
- Bigquery dataset

I think it is a matter of modifying `cookiecutter.json`, perhaps with some additional logic `if AWS then AWS options` and similiar for `if GCP`.

If anyone can point me in the right direction as to which solution is best (quite new to cookiecutter), I can make a PR for this issue.

coleholcomb:
Where are we with this issue? I'd love to pick it up - I want to adopt cookiecutter for my DS team, but we use GCP in our org.

pjbull:
@coleholcomb It's implemented in v2 (#246), which is coming later this year. The only change is in the Makefile for syncing data from the cloud with `gsutil`:
https://github.com/drivendata/cookiecutter-data-science/blob/1fe968d24e73180a9191152d1573dd870994fcae/%7B%7B%20cookiecutter.repo_name%20%7D%7D/Makefile#L40-L64

coleholcomb:
@pjbull got it, will look forward to v2 then, thanks

xzenggit:
Just curious. When will v2 be released?

pjbull:
@cristianmtr For information on progress on v2, see the [branch tracking the overall work](https://github.com/drivendata/cookiecutter-data-science/pull/246). It also has instructions for how to switch over to using v2 as a beta tester, which would be helpful!

jayqi:
Closing—implemented for V2.

Issue #118, opened 2018-05-14
use pipenv?


sliem:
I'm using [pipenv](https://docs.pipenv.org/) a lot which is, and I quote, the officially recommended Python packaging tool from Python.org. Is there interest in supporting pipenv in this cookiecutter?

Something like [this](https://github.com/callumkift/cookiecutter-data-science/commit/b714e0c0b48ee22c127dcbfa78b3f61efb577378) by @callumkift.

pjbull:
I'd love to make pipenv an option. I think that it would need:

 (1) An assessment of how the environment creation would interact with existing `conda` and `vritualenv` options.
 (2) An assessment of how the existing `requirements.txt` based flow would need to change when using the pipenv option.
 (3) It introduces enough complexity that I'd like it to be part of an effort that does a more robust job of running test cases.

Thoughts on any of these? Looks like the option you pointed to is more of a replacement than an option.

callumkift:
Just thought I would quickly chime in. Yes, what I am doing is a replacement, rather than an option. It is much more of a quick replacement rather than a full fleshed out approach, e.g. `pipenv` is only used in the created repo, rather than the overall project.

If this is considered something that the project will move forward on, I would love to help out.

pjbull:
Yeah, I'm interested in determining what would be necessary to make `pipenv` work.

The integration may be a little sticky since we use the existing structure in a number of different ways. For example, we auto-detect `conda`/`virtualenv` for environment creation rather than making a user specify right now. We also feed `requirements.txt` into various commands in the `Makefile`.

If we have a good plan that doesn't result in a lot of end user complexity or code/maintenance complexity, then we'd be open to this.

sliem:
Nice to see interests in pipenv support. I've given it some thoughts, and trying to answers your questions/concerns.

Regarding requirements.txt; pipenv itself do not use requirements.txt but can produce one when required (using pipenv lock -r). It's used in one place in the Makefile (in requirements).

The current Makefile assumes the user want to use conda environment if it is available which might not be true. I would stop auto-detections and promote it to an choice variable in cookiecutter.json:

	"virtualenv": ["pipenv", "conda", "virtualenv"]

We can then use jinja2 conditionals to implement the choice in the Makefile. E.g.

```
lint:
{% if cookiecutter.virtualenv == 'pipenv' -%}
	pipenv run $(PYTHON_INTERPRETER) -m flake8 src
{% else -%}
[...]

PR #162, opened 2019-03-22
[WIP] New version with cleaner options

pjbull:
We've seen a lot of potential features for this where we need to handle forking paths gracefully. By default `cookiecutter` can't do this (see https://github.com/audreyr/cookiecutter/pull/848). It's been years, so we can't reasonably expect this to change upstream...

This implements a monkey-patching workaround to enable this behavior. It introduces a couple of major changes, so here are my recommendations.

Here are the big differences for a user:
 - we now need to run `ccds <path to repo>` instead of `cookiecutter <path to repo>`
 - the options and their defaults are changing
 - as a consequence of supporting more environments/dependency managers at setup time, we stop supporting them at execution time. this means that teams will have to pick one of each and stick to it across developers. I think this is pretty widespread already. (i.e. `make create_environment` will only support one of the options rather than multiple like it does now)

Implementation details are:
 - Add `setup.py` to make this package installable and give it a CLI 
 - add `monkey_patch.py` to patch the `cookiecutter` functions that we need to handle our use case
 - support a list of dictionaries in `cookiecutter.json` that let's a use pick an option and then sub-options
 - use `post_gen_project.py` to create the environment file based on a standard list of libraries. (#5)

There is also work for a number of longstanding items in this branch as well:
 - we now support multiple storage providers (#120)
 - we allow user to choose package manager (#118)
 - we allow user to choose a dependency format
 -  `src` is now `{{ cookiecutter.module_name }}` (#140)
 -  remove `test_environment.py` which is just cruft IMO 

Recommended path forward:
 - [x] tag current master as `v0.1` so anyone relying on the current flow/structure can continue to use it easily
 - [ ] add documentation with new installation (`pip install cookiecutter-data-science`) and new initiation (`ccds <path to repo>`)
 - [x] implement the rails for the high priority items below
 - [x] make comprehensive tests run on CI/CD to de-risk
 - [ ] add table of options with links to the project documentation in our docs
 - [ ] update screencast and add screenshots ( #197 )

I think this also unblocks a whole suite of items that may require choice-specific configuration:
 - choose and configure DAG tool (#156)
 - choose docs package and integration (#7)
 - VCS choice and integration (#56)
 - data version controls (`dvc` or `mlflow`) integration
 - experiment management integration
 - docker support (#62)


docs/docs/index.md
```diff
@@ -72,7 +72,7 @@ cookiecutter https://github.com/drivendata/cookiecutter-data-science
 
 ### Example
 
-<script type="text/javascript" src="https://asciinema.org/a/9bgl5qh17wlop4xyxu9n9wr02.js" id="asciicast-9bgl5qh17wlop4xyxu9n9wr02" async></script>
+<script type="text/javascript" {{ cookiecutter.module_name }}="https://asciinema.org/a/9bgl5qh17wlop4xyxu9n9wr02.js" id="asciicast-9bgl5qh17wlop4xyxu9n9wr02" async></script>
```
jamesmyatt: I think this change is wrong

docs/docs/index.md
```diff
@@ -129,7 +129,7 @@ There are some opinions implicit in the project structure that have grown out of
 
 ### Data is immutable
 
-Don't ever edit your raw data, especially not manually, and especially not in Excel. Don't overwrite your raw data. Don't save multiple versions of the raw data. Treat the data (and its format) as immutable. The code you write should move the raw data through a pipeline to your final analysis. You shouldn't have to run all of the steps every time you want to make a new figure (see [Analysis is a DAG](#analysis-is-a-dag)), but anyone should be able to reproduce the final products with only the code in `src` and the data in `data/raw`.
+Don't ever edit your raw data, especially not manually, and especially not in Excel. Don't overwrite your raw data. Don't save multiple versions of the raw data. Treat the data (and its format) as immutable. The code you write should move the raw data through a pipeline to your final analysis. You shouldn't have to run all of the steps every time you want to make a new figure (see [Analysis is a DAG](#analysis-is-a-dag)), but anyone should be able to reproduce the final products with only the code in `{{ cookiecutter.module_name }}` and the data in `data/raw`.
```
jamesmyatt: Suggest using semantic line breaks: https://rhodesmill.org/brandon/2012/one-sentence-per-line/

setup.py
```diff
@@ -0,0 +1,53 @@
...
+        # that you indicate whether you support Python 2, Python 3 or both.
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+
+        'Topic :: Scientific/Engineering',
+        'Topic :: Scientific/Engineering :: Artificial Intelligence',
+    ],
+    python_requires='>=3.5',
```
jamesmyatt: There's lots of great stuff in Python 3.6, like f-strings, that I'd make that the default minimum version. Or otherwise make the latest version of Python the default.

ned2:
I would suggest changing `cookiecutter.module_name` to `cookiecutter.package_name`. While it's technically true that all Python packages are modules, when arranged into a collection of other modules and having a `__init__.py`, they then also become a package, and this is more salient from the end user perspective. For example, the status as package informs the user that it can be pip installed.

jamesmyatt:
> I would suggest changing `cookiecutter.module_name` to `cookiecutter.package_name`. While it's technically true that all Python packages are modules, when arranged into a collection of other modules and having a `__init__.py`, they then also become a package, and this is more salient from the end user perspective. For example, the status as package informs the user that it can be pip installed.

`module_name` is correct. For example, "sklearn" is the module name for the package named "scikit-learn".

ned2:
> `module_name` is correct. For example, "sklearn" is the module name for the package named "scikit-learn".

Ah yes. Good point. Just found it a little counter-intuitive at first.

tim-werner:
Are there any updates regarding this pull request? The master branch is quite outdated and I think this pull request adds a lot of value and addresses many open issues.

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -36,45 +29,49 @@ clean:
...
-sync_data_from_s3:
-ifeq (default,$(PROFILE))
-	aws s3 sync s3://$(BUCKET)/data/ data/
-else
-	aws s3 sync s3://$(BUCKET)/data/ data/ --profile $(PROFILE)
-endif
-
+	flake8 {{ cookiecutter.module_name }}
+
+{% if cookiecutter.dataset_storage.s3 %}
```
hoangphan1509: should be `{%` if 's3' in cookiecutter.dataset_storage %}`

hooks/post_gen_project.py
```diff
@@ -0,0 +1,66 @@
...
+    'setuptools',
+    'wheel',
+]
+
+pip_only_packages = [
+    'awscli',
+    'python-dotenv',
+]
+
+{% if cookiecutter.dataset_storage.s3 %}
```
hoangphan1509: `{% if 's3' in cookiecutter.dataset_storage %}`

{{ cookiecutter.repo_name }}/Makefile
pjbull: Did you test this? These lines work for me as is.

hoangphan1509: The error only happens when you chose none option in dataset_storage.

hackalog:
Or perhaps you can expect this to change upstream: New maintainier. Cookiecutter 2.0 coming. https://github.com/cookiecutter/cookiecutter/issues/1256

pjbull:
I've got a decent head of steam on a comprehensive set of tests we needed anyway for this change (and for the project generally). I'll keep going to get all the environment options tested on all platforms and then we can assess what to look to cookiecutter for and what to do ourselves.

The biggest decision is whether this becomes a tool where you have to do `ccds ....` or `cookiecutter ...`. I guess, once this is all working, we can always integrate upstream cookiecutter functionality and remove the monkey_patch. However, I wouldn't want to go back to `cookiecutter ...` being the entrypoint since that is annoying churn in the command that you run and the packages you need installed to create environments.

pjbull:
Moving this work on to an official `v2` branch!

Issue #168, opened 2019-05-23
Notebooks from environments cannot find src


jraviotta:
Related to #143 #76 

<details><summary>Configuration</summary>
WSL

Name                    Version                   Build  Channel
alabaster                 0.7.12                   pypi_0    pypi
arrow                     0.13.2                   py36_0    conda-forge
asn1crypto                0.24.0                py36_1003    conda-forge
attrs                     19.1.0                     py_0    conda-forge
awscli                    1.16.164                 pypi_0    pypi
babel                     2.6.0                    pypi_0    pypi
backcall                  0.1.0                      py_0    conda-forge
binaryornot               0.4.4                      py_1    conda-forge
bleach                    3.1.0                      py_0    conda-forge
botocore                  1.12.154                 pypi_0    pypi
bzip2                     1.0.6             h14c3975_1002    conda-forge
ca-certificates           2019.3.9             hecc5488_0    conda-forge
certifi                   2019.3.9                 py36_0    conda-forge
cffi                      1.12.3           py36h8022711_0    conda-forge
chardet                   3.0.4                 py36_1003    conda-forge
click                     7.0                        py_0    conda-forge
colorama                  0.3.9                    pypi_0    pypi
conda                     4.6.14                   py36_0    conda-forge
conda-env                 2.6.0                         1    conda-forge
cookiecutter              1.6.0                 py36_1000    conda-forge
coverage                  4.5.3                    pypi_0    pypi
cryptography              2.6.1            py36h72c5cf5_0    conda-forge
cryptography-vectors      2.6.1                      py_0    conda-forge
cycler                    0.10.0                   pypi_0    pypi
dbus                      1.13.6               he372182_0    conda-forge
decorator                 4.4.0                      py_0    conda-forge
defusedxml                0.5.0                      py_1    conda-forge
docutils                  0.14                     pypi_0    pypi
entrypoints               0.3                   py36_1000    conda-forge
expat                     2.2.5             hf484d3e_1002    conda-forge
flake8                    3.7.7                    pypi_0    pypi
fontconfig                2.13.1            he4413a7_1000    conda-forge
freetype                  2.10.0               he983fc9_0    conda-forge
future                    0.17.1                py36_1000    conda-forge
gettext                   0.19.8.1          hc5be6a0_1002    conda-forge
glib                      2.58.3            hf63aee3_1001    conda-forge
gmp                       6.1.2             hf484d3e_1000    conda-forge
[...]

PR #170, opened 2019-05-23
Added support for auto-creation of conda environments

jraviotta:
This is one possible resolution for #168 .

This PR updates the `Makefile requirements` target with logic for selecting package and environment managers. If conda is detected, a conda environment is created and then updated with a new `environment.yml` file. If conda is not detected, `requirements.txt` is installed with pip and then an environment is created with virtualenv as before.

The conda condition successfully creates the environment and the local package `src` is available in the jupyter kernel in jupyter lab. I can run the notebook cell described in #168 without an error. 

I have not tested the virtualenv condition as I am unfamiliar with it's normal behavior.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,17 @@
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - conda
+  - click
+  - Sphinx
+  - coverage
+  - awscli
+  - flake8
+  - python-dotenv
+  - setuptools
+  - wheel
+  - pip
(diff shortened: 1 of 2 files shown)


pjbull:
Thanks for the interest in this feature @jraviotta!

We're not going to merge this as-is, since we're moving to allowing the user to pick an option rather than bundling everything together. (see #162) If you want to test the changes on that branch with your `conda` flow, it would be great to hear the results.

PR #187, opened 2019-10-15
add tests folder and basic test

tim-werner:
Add test folder and basic test to the cookiecutter template.


jamesmyatt:
When adding tests outside of the application, it's usual to either avoid making tests a module or to use a [`src` layout](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure). See https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code

pjbull:
Closing since I think we will reconsider options for this after #162 is in. Issue filed #195

PR #194, opened 2020-02-08
Set up CI with Azure Pipelines

drivendata:


pjbull:
Working on this on top of #162 instead so that the new comprehensive tests there can get run against all of the platforms.

jayqi:
So the only remaining failing tests are for `virtualenvwrapper` on Windows. Tried a bunch of things and not able to get them working. The gist of the problem is that `which virtualenvwrapper.sh` does not work. `virtualenvwrapper` is installed, and the directory containing that script is in `$PATH`, but it does not turn up with `which` because it's not executable. I've tried the following:
- Using `find` to search all `$PATH` directories to grab it.
   - Sourcing it directly helps, but there's a ``source `which virtualenvwrapper.sh` `` in the Makefile as well. 
   - Not able to `chmod` it to be executable because I guess that's a Windows file system limitation. 
   - Adding a `#!/bin/bash` shebang to the first line (which it doesn't have); I tested this on the AWS Workspace machine, and this worked there in making Git Bash recognize the file as an executable. Don't know why it doesn't work on Azure Pipelines. 
- Tried installing [`virtualenvwrapper-win`](https://pypi.org/project/virtualenvwrapper-win/). That didn't seem to do anything.

I think what I would recommend is using `find` to grab the path of `virtualenvwrapper.sh` and sourcing that, and then removing the ``source `which virtualenvwrapper.sh` `` from Makefile. That would require users to have properly sourced it in the `.bash_profile` or whatever, instead of CCDS doing it for them.

PR #196, opened 2020-02-11
Fixed test_folders test to check for new folders in repository

ESKYoung:
Changed from set difference to set symmetric difference. Test will now fail if there are more folders in the repository than listed in the expected_dirs variable.


pjbull:
Thanks @ESKYoung. I'm actually rewriting the test suite in #162 right now where direct list comparisons in pytest give symmetric difference. Good catch though. Closing for now.

PR #199, opened 2020-02-19
handle multiple data providers

ianepreston:
Hopefully this is the right place for this, apologies if not. This is my first attempt contributing to this project. I noticed in a comment in #162 that ```commands.rst``` doesn't build with the new cloud provider options. I think this resolves the issue, but it's hard for me to test completely as I hit later errors in the project. This does seem to resolve that specific error at least.

If there's any additional work I can do to help with the merge process, or if there's a better place to submit this please let me know.

Thanks


pjbull:
Thanks!

PR #204, opened 2020-03-11
Add gh setup

bradyrx:


bradyrx:
I'm sorry, I keep doing this accidentally. I should see if I can change it to default to my local fork somehow.

PR #205, opened 2020-03-19
Remove source `which virtualenvwrapper.sh` from Makefile

jayqi:
`which virtualenvwrapper.sh` fails on Windows. I propose we remove it from the Makefile command. This means that CCDS assumes a user has virtualenvwrapper set up properly in their shell startup configuration.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -95,7 +95,6 @@ jobs:
       python -m pip install --upgrade pip
       pip install -r requirements.txt
       pip install -r dev-requirements.txt
-      pip install virtualenvwrapper-win
       pip install -e .
     displayName: 'Install dependencies'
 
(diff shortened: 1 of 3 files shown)


omri374:
`which` is a unix command. In windows the same command is `where`. Maybe this can help? https://gist.github.com/sighingnow/deee806603ec9274fd47

pjbull:
>which is a unix command. In windows the same command is where. Maybe this can help? https://gist.github.com/sighingnow/deee806603ec9274fd47

Pretty sure that the `Bash task` on Azure Pipelines will run our code in a bash-like environment across all operating systems so that shouldn't be the issue
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/bash?view=azure-devops

pjbull:
Closing this since it was fixed in #217

PR #211, opened 2020-04-22
Tox reference

natravedrova:
Use https://tox.readthedocs.io/ instead of insecure and somehow confused  tox.testrun.org.

This was already mentioned in the issues and it looks like there was respective pull request, but for a reason I don't see better link to the tox project.


natravedrova:
Related issue: #10

pjbull:
Can you grab the other reference in docs/docs/index.md too?

pjbull:
Thanks!

PR #216, opened 2020-05-24
Fixed failing tests on Windows

### LINKED ISSUE #215: Repo tests are failing on Windows
Some tests fail on `py.test tests` due to:
```
TypeError: argument of type 'WindowsPath' is not iterable
```

To reproduce:
1. Clone repo
2. Run `pip install -r requirements.txt`
3. Run `py.test tests`

This is a Python pathlib bug which is only fixed on Python 3.8: https://github.com/python/cpython/pull/7071

However a possible workaround is to change implementation to not iterate through the path.

Full test output:
```
self = <test_creation.TestCookieSetup object at 0x00000144B1B37550>

    def test_author(self):
        setup_ = self.path / 'setup.py'
        args = ['python', setup_, '--author']
>       p = check_output(args).decode('ascii').strip()

tests\test_creation.py:38:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
c:\programdata\miniconda3\Lib\subprocess.py:395: in check_output
    **kwargs).stdout
c:\programdata\miniconda3\Lib\subprocess.py:472: in run
    with Popen(*popenargs, **kwargs) as process:
c:\programdata\miniconda3\Lib\subprocess.py:775: in __init__
    restore_signals, start_new_session)
c:\programdata\miniconda3\Lib\subprocess.py:1119: in _execute_child
    args = list2cmdline(args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

seq = ['python', WindowsPath('C:/Users/user/AppData/Local/Temp/pytest-of-user/pytest-7/data-project0/DrivenData/setup.py'), '--author']

    def list2cmdline(seq):
        """
        Translate a sequence of arguments into a command line
        string, using the same rules as the MS C runtime:

        1) Arguments are delimited by white space, which is either a
           space or a tab.

        2) A string surrounded by double quotation marks is
           interpreted as a single argument, regardless of white space
           contained within.  A quoted string can be embedded in an
           argument.

        3) A double quotation mark preceded by a backslash is
           interpreted as a literal double quotation mark.

        4) Backslashes are interpreted literally, unless they
           immediately precede a double quotation mark.

        5) If backslashes immediately precede a double quotation mark,
           every pair of backslashes is interpreted as a literal
           backslash.  If the number of backslashes is odd, the last
           backslash escapes the next double quotation mark as
           described in rule 3.
        """

        # See
        # http://msdn.microsoft.com/en-us/library/17w5ykft.aspx
        # or search http://msdn.microsoft.com for
        # "Parsing C++ Command-Line Arguments"
        result = []
        needquote = False
        for arg in seq:
            bs_buf = []

            # Add a space to separate this argument from the others
            if result:
                result.append(' ')

>           needquote = (" " in arg) or ("\t" in arg) or not arg
E           TypeError: argument of type 'WindowsPath' is not iterable
```

omri374:
Casting Path objects to strings due to an issue with `WindowsPath`:
```
TypeError: argument of type 'WindowsPath' is not iterable
```

Fixes #215

### DIFF
--- a/tests/test_creation.py
+++ b/tests/test_creation.py
@@ -8,15 +8,10 @@ def no_curlies(filepath):
     """ Utility to make sure no curly braces appear in a file.
         That is, was Jinja able to render everything?
     """
-    with open(filepath, 'r') as f:
+    with open(filepath, "r") as f:
         data = f.read()
 
-    template_strings = [
-        '{{',
-        '}}',
-        '{%',
-        '%}'
-    ]
+    template_strings = ["{{", "}}", "{%", "%}"]
 
     template_strings_in_file = [s in data for s in template_strings]
     return not any(template_strings_in_file)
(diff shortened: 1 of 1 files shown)


omri374:
@pjbull Done

PR #220, opened 2020-08-28
updated .gitignore

kd2718:
On complicated projects, I used to get confused on what I added to the .gitignore and what was recommended. There was times that I might want to remove something that I previously added. 

Adding some basic annotations helped me keep track of what I added and what was set in place by the recommended .gitignore file.


pjbull:
@kd2718 Appreciate the sentiment, but I know that there is a huge diversity in how people manage their `.gitignore` files. (e.g., by tool rather than user/default). I think we'll leave that open-ended rather than recommend splitting into user/default.

PR #223, opened 2020-10-02
Better gitignore for data directory

jamesmyatt:


qalgit:
Lgtm.

{{ cookiecutter.repo_name }}/data/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+**/*
+!**/
+!/*
+
+# Include git files
+!.gitignore
+!.gitkeep
+
+# Include DVC files
+!*.dvc
```
jayqi, line 13: I'm not sure I like putting a .gitignore inside `data/`. I think it's better to keep these rules in the repo-root-level `.gitignore` and not mix this into the data directory. This is going to end up getting caught up syncing to the cloud, etc.

{{ cookiecutter.repo_name }}/data/.gitignore
```diff
@@ -0,0 +1,13 @@
+# Ignore all files in all sub-directories by default
+# but not the sub-directories themselves
+# and not files in this directory
+**/*
+!**/
+!/*
```
jayqi, line 6: Git only tracks files and not directories, so there's no point in trying to have an include rule for directories.

jayqi (changes requested):

PR #228, opened 2020-12-01
Simplify cookiecutter

ivyleavedtoadflax:
Simplifies the `cookiecutter-data-science` template to something a bit more like the development environment we actually use:

* Adds a makefile which creates a virtualenv with venv
* Moves to a system of unpinned requirements that are pinned with `make update-requirements`
* Adds a .envrc file for use with direnv and sets aws profile there
* Splits source dependencies and environment dependencies into requirements_dev.txt and requirements_dev.txt
* Other minor updates


ivyleavedtoadflax:
My bad! Wrong repo!

PR #236, opened 2020-12-31
Added setup.py to project organization in README

bernardoamaral1011:
Small fix to the top level README since it was not coherent with the rest of the project and documentation.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ The directory structure of your new project looks like this:
 ├── requirements.txt   <- The requirements file for reproducing the analysis environment, e.g.
 │                         generated with `pip freeze > requirements.txt`
 │
+├── setup.py           <- makes project pip installable (pip install -e .) so src can be imported
 ├── src                <- Source code for use in this project.
 │   ├── __init__.py    <- Makes src a Python module
 │   │


qalgit:
@bernardoamaral1011,

Looking good, some nice to haves....:

1. **Consistency**: Some sentences end with a "." and others don't. Nice to make closure with all sentences using a "."
2. **notebooks** folder : Should there be another folder = jupyter-notebooks under the generic **notebooks** folder ? As there are other alternative Jupyter notebooks such as R notebooks etc.

Cheers,
Laurence.

bernardoamaral1011:
Hi there!

This PR is just a typo fix, so that README.md is coherent with {{ cookiecutter.repo_name }}/README.md, as well as with the directory structure of the template.
However, I do agree with the first suggestion.

The second suggestion is out of the scope of this PR and an issue should probably be opened to discuss it.

jamesmyatt:
@qalgit, for #2, I'd say no. Use the "notebooks" directory for any kind of notebook. So if your project uses Jupyter notebooks, then put the ipynb files there. If your project uses R Markdown notebooks, then put the Rmd files there. If your project uses jupytext, then put the scripts there. Doesn't matter.

The only change I'd consider making in this regard is to remove the "Juypter" part of the description. But it's up to each project to customise the repo for their purposes after it's been generated anyway. In my projects, I always rename the directory to "analysis" so that the scope is clearer, and change the convention to remove the author's initials.

pjbull:
Thanks @bernardoamaral1011!

On the consistency point, we'd welcome an issue where we can draft consistent language.

On the notebooks folder, we won't add another level for different languages, that is way too uncommon a use case.

PR #242, opened 2021-03-03
Docker enhancement

alejandrobalderas:
Added a Dockerfile and docker-compose.yml. 

By running `docker-compose up` you can launch jupyter lab within the docker container. This allows you to develop your analysis in a containerized environment that you can later share or deploy. Developing inside the container ensures you that your environment will be the same that you can use in production later on. 

I had issues where even using a requirments.txt I could not get the exact same build as some other colleges. I have been using this technique a while now and I wanted to share it with you guys ;) 

Happy to hear your comments about it

### DIFF
--- a/{{ cookiecutter.repo_name }}/Dockerfile
+++ b/{{ cookiecutter.repo_name }}/Dockerfile
@@ -1,27 +1,23 @@
 FROM python:3.9-slim
 
-RUN apt update && apt install -y \
-    sudo
 
-ARG USERNAME=dev
-ARG USER_UID=1000
-ARG USER_GID=$USER_UID
+# Gets all updates if needed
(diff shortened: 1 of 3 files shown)


pjbull:
It's not currently on our roadmap to adopt Docker (there is a bunch of discussion in issues on that). It's our current opinion that it is too much overhead to make the default for CCDS since our user base is so broad (students, academics, industry teams, freelancers, data journalists, etc.)

This also doesn't seem like we need to solve it because some downstream projects have done a nice job already of making it work for the DS team use case (for example, https://github.com/manifoldai/orbyter-cookiecutter ).

Thanks for the thoughts, though. We will revisit this as we develop CCDS v2, which may optionally support Docker.

PR #247, opened 2021-04-02
Lint, format, license

tamara-glazer:
- Adds formatting commands (`black`) and pre-commit wrapper
- Configuration file format defaults (`setup.cfg` and `pyproject.toml`)
- Support Python 3.8, remove 3.5
- Default to no license (rather than MIT)

### DIFF
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean data lint requirements sync_data_down sync_data_up
+.PHONY: clean data lint format requirements sync_data_down sync_data_up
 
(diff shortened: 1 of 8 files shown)


{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -82,6 +86,10 @@ create_environment:
 	{% endif %}
 {% endif %}
 
+## Wrapper for format and lint
+pre-commit:
+	make format
+	make lint
```
pjbull, line 92: I'd cut this since we're VCS agnostic

pjbull (commented):
Thanks @tamara-glazer these are great!

PR #251, opened 2021-07-20
Minor docs fix: Fix broken joblib link

carderne:
Joblib link seems to be dead. Not sure what your preferred PR flow is but feel free to ignore and include in your next docs update. :)


pjbull:
Fixed in V2, thanks for reporting!

PR #270, opened 2022-06-13
Update index.md

nitinramvelraj:
Requirements state python version 2.7 or 3.5, I believe it should be 3.5+


pjbull:
Updated in V2, thanks for reporting!

PR #275, opened 2022-07-27
Adding the ability to sort imports in src using isort

arturomoncadatorres:
This tiny PR adds a rule to the `Makefile` to sort the imports of `src` using [`isort`](https://github.com/PyCQA/isort).


pjbull:
@arturomoncadatorres I think we'd consider this during the transition to v2. Could you file an issue instead and link to your branch in your fork?

arturomoncadatorres:
Just did ;) !

PR #286, opened 2022-10-13
Optionally remove boilerplate in initial setup

### LINKED ISSUE #49: Make Python boilerplate optional in initial setup
This makes the project cleaner for other languages, e.g. R, Julia, etc.

chrisjkuch:
Closes #49 

Provides additional prompt to make boilerplate python module files and folders disappear, with the exception of `__init__.py` to make the module a python module.

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -1,6 +1,5 @@
-from pathlib import Path
 import shutil
-
+from pathlib import Path
 
 # https://github.com/cookiecutter/cookiecutter/issues/824
 #   our workaround is to include these utility functions in the CCDS package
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -40,6 +41,10 @@ def config_generator(fast=False):
         ],
         [("dependency_file", opt) for opt in cookiecutter_json["dependency_file"]],
         [("pydata_packages", opt) for opt in cookiecutter_json["pydata_packages"]],
+        [
+            ("include_code_scaffold", opt)
+            for opt in cookiecutter_json["include_code_scaffold"]
+        ],
```
chrisjkuch, line 47: This doesn't need to be in the `product`

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -37,11 +36,13 @@ clean:
 lint:
 	flake8 {{ cookiecutter.module_name }}
 	black --check --config pyproject.toml {{ cookiecutter.module_name }}
+	isort --check --profile black {{ cookiecutter.module_name }}
```
chrisjkuch, line 39: Let's not do this as part of this PR

pjbull (approved):
Looks good, thanks @chrisjkuch!

hooks/post_gen_project.py
```diff
@@ -14,12 +15,21 @@
     "isort",
     "pip",
     "python-dotenv",
+    "setuptools",
+    "wheel",
```
jayqi, line 19: @chrisjkuch why do we need setuptools and wheel? These should be unnecessary for a pyproject.toml-based build (this stuff gets specified in pyproject.toml).

PR #287, opened 2022-10-13
Updating the Makefile

AllenDowney:
Adding create-environment command, global variables, and a few comments (making the Makefile for this project more like the Makefile it generates)


pjbull (approved):
Looks great!

PR #288, opened 2022-10-13
Removing deprecation message from v2

### LINKED ISSUE #281: Running v2 should not generate a v1 deprecation warning
I installed and ran the current draft of v2 like this:

```
$ pip install git+https://github.com/drivendata/cookiecutter-data-science@v2
$ ccds https://github.com/drivendata/cookiecutter-data-science.git  --checkout v2
```

After a successful run, I still get the deprecation message about v1.

AllenDowney:
Removing the deprecation message from the pre-rendering hook (and since there's nothing left, adding a pointer to the documentation)

Closes #281

### DIFF
--- a/hooks/pre_gen_project.py
+++ b/hooks/pre_gen_project.py
@@ -1,24 +1,4 @@
-def deprecation_warning():
-    print(
-        """
+# Functions here run before the project is generated.
 
-=============================================================================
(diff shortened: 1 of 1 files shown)


pjbull (approved):
Thanks!

PR #295, opened 2022-11-04
Replace sphinx with mkdocs as default docs option, add none option

### LINKED ISSUE #327: Switch from sphinx as default project docs to mkdocs
Mkdocs is a bit easier / lighter weight.

chrisjkuch:
Closes #327 

This PR allows a ccds user to select mkdocs or none as their code documentation tool. mkdocs is the new default.

If `mkdocs` is chosen, provides the output of `mkdocs new {{ project_name }}` in `/docs/`

If `none` is chosen, provides no default documentation tool, but keeps an empty `/docs/` folder.

### Notes

To prevent an explosion of possible configuration options in the `product` that generates test configs, this PR also switches to simply cycling through all possible values for multi-valued configuration options that should be nominally independent (storage backend, license, including the code scaffold, and docs tool).


jayqi:
@chrisjkuch do you have a sense of where this PR is at? Having a refreshed design around docs seems like it may be worth including in V2. 

I was chatting with @pjbull and the desired goal may not be exactly what this was originally doing though. Given that markdown is much more likely to be useful that RST, it probably makes more sense to _replace_ Sphinx with mkdocs, rather than introducing complexity to have Sphinx as an option. Furthermore, it seems also likely that many CCDS-derived projects don't actually need a docs website, and so it may make sense for the choices to be None or mkdocs.

chrisjkuch:
This PR is definitely out of date. I agree that straight-up replacement of Sphinx with mkdocs is best (and I think is captured by #327), while also providing a "none" option (which isn't currently captured in that issue). For ease, I might close this out and create a new PR, since this looks like a bit of a git mess.

chrisjkuch:
Upon further reflection, it would be more effort to remove sphinx than to leave it in. If we have a strong desire to remove it, I can do so, but the easiest thing is to leave it as a non-default option.

{{ cookiecutter.repo_name }}/docs/mkdocs/docs/index.md
```diff
@@ -16,4 +17,13 @@ Syncing data to cloud storage
...
+{% endif %}
+
+
+## Project layout
+
+    mkdocs.yml    # The configuration file.
+    docs/
+        index.md            # The documentation homepage.
+        getting-started.md  # Set-up instructions
+        ...       # Other markdown pages, images and other files.
```
pjbull, line 29: The docs here are for the whole project, not just for the documentation itself, so maybe this doesn't belong here?

hooks/post_gen_project.py
```diff
@@ -47,6 +47,18 @@
...
+# or none if none selected
+docs_path = Path("docs")
+# {% if cookiecutter.docs != "none" %}
+packages += ["{{ cookiecutter.docs }}"]
+pip_only_packages += ["{{ cookiecutter.docs }}"]
+docs_subpath = docs_path / "{{ cookiecutter.docs }}"
+for obj in docs_subpath.iterdir():
+    shutil.move(str(obj), str(docs_path))
+# {% endif %}
+shutil.rmtree(docs_path / "mkdocs")
```
pjbull, line 60: Maybe should be a for loop that removes all the other options for `cookiecutter.docs` that were not selected if they exist so we have a little future proofing.

pjbull (commented):
A couple little things, thanks @chrisjkuch

{{ cookiecutter.repo_name }}/docs/mkdocs/docs/index.md
chrisjkuch: Great point. Would a `tree` of the high-level default structure be appropriate? We can also just delete.

PR #297, opened 2022-12-13
make gsutil rsync recursive

KBodolai:
Small change to allow gsutil to do rsync recursively, syncing all the folders.

Context: We had some issues using it with gcloud, since `make sync_data_up` would not upload all the data folders. Adding the -r flag fixes this issue.


pjbull:
Thanks @KBodolai. Rebasing on to the latest `v2` should work now to run the tests.

KBodolai:
ok, I've done that, hopefully it'll be it!

KBodolai:
ok, so these passed (yay!) 

There's another parameter that can be fed to gsutil for threaded processing (-m), which provides huge performance boosts, although they mention [in some cases it may slow the syncing down](https://cloud.google.com/storage/docs/gsutil/commands/rsync).

I believe it will typically only result in speedups for the setting this is used in, but not quite certain, what do you think?

pjbull:
@KBodolai I agree that 99% of the use cases here will be cloud <-> local, not local <-> local so `-m` likely makes sense. Happy to wait on this til that id added

KBodolai:
right, I'll add it and set up a test project to double check it does well, I'll ping you when it's ready :)

KBodolai:
@pjbull , actually, I stayed a bit late after work and tested if with a bunch of heavy lidar files, all working fine!

PR #309, opened 2023-07-03
Fixed link to Twelve Factor App

jhgee:
http is not redirected -> changed link to https


pjbull:
Fixed in V2, thanks for reporting!

PR #335, opened 2023-12-01
Update Makefile help command to work on all platforms

### LINKED ISSUE #221: Solo `make` command doesn't usually work (generate help text) on Windows
Encountered this in a downstream project:
https://github.com/drivendataorg/sfp-cervical-biopsy-runtime/issues/18

Fix was to just remove the `more` command since it's not available on most windows bash shells:
https://github.com/drivendataorg/sfp-cervical-biopsy-runtime/pull/19

Expected behavior: running `make` generates help text

- [x] ensure help text works as expected
- [ ] #333
- [x] consider adding test case
  - [x] for make generating help text (or at least, not failing on the github actions windows runner)

chrisjkuch:
Closes #221

This PR switches the style and format of the python script in the Makefile that generates the help message text. The old script was not formatted correctly to regex match commented commands written with the structure we use in our Makefile, and leveraged `$$VAR` syntax in the command itself that does not work in Windows environments.

In order for this to work in the intended way, we had to treat the Python script as a single line - otherwise, because we're invoking the script as a command, newlines would trigger running a new command entirely. That's why the script ends each line with `; \`.

### DIFF
--- a/tests/conda_harness.sh
+++ b/tests/conda_harness.sh
@@ -29,6 +29,7 @@ then
     sudo chown -R $USER /usr/local/miniconda
 fi
 
+make
 make create_environment
 conda activate $PROJECT_NAME
 make requirements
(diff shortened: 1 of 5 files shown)


chrisjkuch:
@pjbull this is ready for review!

pjbull (approved):

Issue #334, opened 2023-12-01
Make v1 template docs accessible in new docs


klwetstone:
- [ ] add docs versioning so that ~there is a dropdown similar to [cloudpathlib](https://cloudpathlib.drivendata.org/~latest/)~ v1 docs show up in a page on the new docs site

See the cloudpathlib [`release.yml`](https://github.com/drivendataorg/cloudpathlib/blob/945a352e0264be66057f1a51b4f25b965b566973/.github/workflows/release.yml#L40) for reference.

Issue #315, opened 2023-08-31
Transfer ownership from drivendata user to drivendataorg organization


jayqi:
Post v2, pre v2.1

- [x] Update GitHub pages website to redirect to new website (follow [this procedure](https://github.com/isaacs/github/issues/492#issuecomment-385309053)) -- see comment below
- [x] Transfer ownership
- [x] Update links in README, etc. for V2

Don't change:

- V1 README instructions to point to new repo: GitHub should redirect this anyways
- V1 README links: these should just redirect, and will be replaced for V2

jayqi:
I'm going to reorder this to put the GitHub Pages step first. I think it's probably safer to get that sorted out separately _before_ the transaction. 

The high-level ideas involved are:

1. Create the user-level GitHub Pages website for the drivendata user (https://drivendata.github.io).
2. Stop deploying/disable the drivendata/cookiecutter-data-science GitHub Pages
3. Create a cookiecutter-data-science page on the drivendata user GitHub Pages website, which should end up with the same URL as the old website. Have this new page do HTML meta redirect to the new https://cookiecutter-data-science.drivendata.org/ URL

Make sure that all works.

jayqi:
Just noting here that we needed to "Delete" (not just "Unpublish") the GitHug Pages site before the user Pages version kicked in: https://docs.github.com/en/pages/getting-started-with-github-pages/deleting-a-github-pages-site

PR #336, opened 2023-12-01
Add release GitHub actions

### LINKED ISSUE #317: Add CI for releasing package to PyPI
- [x] Add GitHub actions to publish on a release
- [ ] Add PyPI secrets to repo

### LINKED ISSUE #389: Release tagging mechanics
For changes to defaults, we'll want well structured release mechanics.

This has come up since #387 is proposing changing defaults.

Here's a proposal for release mechanics:
 - Develop feature branches PR'd into `master`
 - Release regularly following semantic versioning
 - Couple both `ccds` code + template changes together in a single version number (maybe we split these out later, but lets keep it simple for now)
 - On release workflow, tag commit with version number
 - Publish corresponding `ccds` package version number to PyPI
 - Default `ccds` clones the template @ the version tag
 - Add `--latest` flag to get the latest from `master` instead

Thoughts?

klwetstone:
closes #317 
closes #389 

## Add GitHub actions to publish on release

~**Do not merge** - [#315](https://github.com/drivendata/cookiecutter-data-science/issues/315) is needed first. After transfer, the repo should automatically have access to the PyPI secrets needed by `release.yml`~

### `ccds.__version__`

Users can now see the version of `ccds` after importing (`ccds.__version__`). However, the *project* name is different than the *module* name. This means that in `version.py`, we have to hard code the project name with:

```python
__version__ = importlib_metadata.version("cookiecutter-data-science")
```

Instead of using:

```python
__version__ = importlib_metadata.version(__name__.split(".", 1)[0])
```

### Using release version as default tag for checkout

Per #389, we wanted a way to allow more stable behavior when someone installs ccds. Currently, any change to the template on the master branch will automatically get propagated when someone runs ccds (unless the person has downloaded the template already and opts to skip that step). As part of this PR, we now default to ccds checking out the tag associated with the installed ccds version. 

Users can always get the latest by checking out the master branch (`-c master`) or updating ccds.

### Docs

This PR does not include versioning for docs (eg. with a "Build docs" job). That is tracked separately in #334

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,14 @@ The directory structure of your new project will look something like this (depen
     └── plots.py                <- Code to create visualizations   
 ```
 
+## Using unreleased changes
+
+By default, `ccds` will download the most recently _released_ version of the template. If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
+
+```bash
+ccds -c master
+```
+
 ## Using v1
 
 If you want to use the old v1 project template, you need to have either the cookiecutter-data-science package or cookiecutter package installed. Then, use either command-line program with the `-c v1` option:
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The directory structure of your new project will look something like this (depen
 
 ## Using unreleased changes
 
-By default, `ccds` will use the version of the template that corresponds to the installed version (e.g., if you have installed version v2.0.1, you'll use the v2.0.1 version of the template by default). If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
+By default, `ccds` will use the _project template_ version that corresponds to the _installed `ccds` package_ version (e.g., if you have installed `ccds` v2.0.1, you'll use the v2.0.1 version of the project template by default). To use a specific version of the project template, use the `-c/--checkout` flag to provide the branch (or tag or commit hash) of the version you'd like to use. For example to use the project template from the `master` branch:
 
 ```bash
 ccds -c master
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The directory structure of your new project will look something like this (depen
 
 ## Using unreleased changes
 
-By default, `ccds` will download the most recently _released_ version of the template. If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
+By default, `ccds` will use the version of the template that corresponds to the installed version (e.g., if you have installed version 2.0.1, you'll use the 2.0.1 version of the template by default). If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
 
 ```bash
 ccds -c master
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The directory structure of your new project will look something like this (depen
 
 ## Using unreleased changes
 
-By default, `ccds` will use the version of the template that corresponds to the installed version (e.g., if you have installed version 2.0.1, you'll use the 2.0.1 version of the template by default). If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
+By default, `ccds` will use the version of the template that corresponds to the installed version (e.g., if you have installed version v2.0.1, you'll use the v2.0.1 version of the template by default). If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
 
 ```bash
 ccds -c master
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,7 +2,7 @@
 
 ## Background
 
-The release of [ccds v2](https://drivendata.co/blog/ccds-v2) introduced the `ccds` utility and the concept of versioning to cookiecutter data science. Prior to this release, there was just the template and the use of [cookiecutter](https://github.com/cookiecutter/cookiecutter) - branches and forks could be used in the usual way to get different versions of the template.
+The release of [ccds v2](https://drivendata.co/blog/ccds-v2) introduced the `ccds` utility and the concept of versioning to cookiecutter data science. Prior to this release, cookiecutter-data-science only provided a project template, which the generic [cookiecutter](https://github.com/cookiecutter/cookiecutter) utility could use to instantiate a project. Branches and forks could be used in the usual way to get different versions of the template.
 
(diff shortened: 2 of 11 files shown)


ccds/__main__.py
```diff
@@ -31,6 +33,9 @@ def _main(*args, **kwargs):
         f.params[1].default = (
             "https://github.com/drivendataorg/cookiecutter-data-science"
         )
+        # The fifth parameter is the "checkout" option in the cookiecutter cli
+        # Per #389, set this to the currently released version by default
+        f.params[4].default = __version__
```
pjbull, line 38: I feel like first param being the repo is unlikely to change, but I am a little worried about ones later in the ordering. Is there a way we can loop over parameters and check by name to find the right one?

chrisjkuch: This is updated to find the "checkout" option by name in the list of cookiecutter parameters.

docs/docs/all-options.md
```diff
@@ -3,4 +3,12 @@
 CCDS provides a number of choices that you can use to customize your project. The defaults work well for many projects, but lots of tooling choices are supported. Here are the options for tools that you can use:
 
 
-<!-- configuration-table.py output -->
\ No newline at end of file
+<!-- configuration-table.py output -->
+
+## Checking out other branches / using unreleased changes to the template
+
+By default, `ccds` will download the most recently _released_ version of the template. If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
```
jayqi, line 10: Is this still accurate? This seems like it defaults to match the current version of `ccds` that is installed.

chrisjkuch: Good catch, it's not, I'll update.

ccds/__main__.py
```diff
@@ -31,6 +33,11 @@ def _main(*args, **kwargs):
         f.params[1].default = (
             "https://github.com/drivendataorg/cookiecutter-data-science"
         )
+        # Find the "checkout" option in the cookiecutter cli (currently the fifth)
+        # Per #389, set this to the currently released version by default
+        param_names = [p.name for p in f.params]
+        checkout_index = param_names.index("checkout")
+        f.params[checkout_index].default = __version__
```
pjbull, line 40: It looks like some places (like the release tags), we use `vX.Y.Z` and others we use `X.Y.Z`. We should check if we need the `v` prefix here or not.

pjbull (commented):
Great! Looking good with one question.

May be nice to also add some instructions on creating a release for maintainers so they know to tag the release in the right way.

ccds/__main__.py
chrisjkuch: @pjbull great catch, would have failed as-is without the "v" prefix. I added a `RELEASING.md` with some instructions, but feel free to tweak or make suggestions.

README.md
```diff
@@ -82,6 +82,14 @@ The directory structure of your new project will look something like this (depen
     └── plots.py                <- Code to create visualizations   
 ```
 
+## Using unreleased changes
+
+By default, `ccds` will use the version of the template that corresponds to the installed version (e.g., if you have installed version v2.0.1, you'll use the v2.0.1 version of the template by default). If there are any _unreleased_ changes to the template (or changes in a separate branch) that you want to incorporate, you can do so by checking out whatever branch you'd like to use (checkout `master` for the latest changes):
```
r-b-g-b, line 87:
```suggestion
By default, `ccds` will use the _project template_ version that corresponds to the _installed `ccds` package_ version (e.g., if you have installed `ccds` v2.0.1, you'll use the v2.0.1 version of the project template by default). To use a specific version of the project template, use the `-c/--checkout` flag to provide the branch (or tag or commit hash) of the version you'd like to use. For example to use the project template from the `master` branch:
```

Seems to be referred to as "project template" elsewhere (easier for me to comprehend in any case).

RELEASING.md
```diff
@@ -0,0 +1,11 @@
+# Information for releases and versioning of ccds
+
+## Background
+
+The release of [ccds v2](https://drivendata.co/blog/ccds-v2) introduced the `ccds` utility and the concept of versioning to cookiecutter data science. Prior to this release, there was just the template and the use of [cookiecutter](https://github.com/cookiecutter/cookiecutter) - branches and forks could be used in the usual way to get different versions of the template.
```
r-b-g-b, line 5:
```suggestion
The release of [ccds v2](https://drivendata.co/blog/ccds-v2) introduced the `ccds` utility and the concept of versioning to cookiecutter data science. Prior to this release, cookiecutter-data-science only provided a project template, which the generic [cookiecutter](https://github.com/cookiecutter/cookiecutter) utility could use to instantiate a project.  Branches and forks could be used in the usual way to get different versions of the template.
```

Slight rewording -- still accurate?

r-b-g-b (changes requested):
Really nice! A couple of wording tweaks using my status as "kind of still figuring out what's going" on to write what made it click for me.

Also maybe unrelated but I notice that

```bash
ccds --version

## Cookiecutter 2.6.0 from /home/robert/projects/codepath-data-infrastructure/.venv/lib/python3.13/site-packages (Python 3.13.1 (main, Dec 19 2024, 14:32:25) [Clang 18.1.8 ])
```

Outputs the `cookiecutter` version (2.6.0), not the `ccds` version (2.0.1). May be inert for the time being and if so I can file an issue for another day.

RELEASING.md
chrisjkuch: yeah this is much more clear

chrisjkuch:
@r-b-g-b could you take one more look here? Applied your suggestions and added a warning about v2.0.0 per @pjbull, who has otherwise signed off

hooks/pre_prompt.py
```diff
@@ -0,0 +1,12 @@
...
+from ccds.version import __version__ as version
+
+if __name__ == "__main__":
+    if version <= "2.0.0":
+        warnings.warn(
+            "It looks like you're using a version of CCDS that "
+            "always uses the latest template. For more stable behavior, "
+            "update to a newer version of CCDS.",
+            DeprecationWarning,
+        )
```
r-b-g-b, line 12: [nit] my brain found it easier to reason about < 2.0.1 than <= 2.0.0 ("if the version is less than 2.0.1, show the deprecation warning")

Then tweaked the wording a bit -- still correct?

```suggestion
    if version < "2.0.1":
        warnings.warn(
            "It looks like you're using a version of CCDS that "
            "defaults to using the latest template. For more stable behavior, "
            "update to CCDS version 2.0.1 or later.",
            DeprecationWarning,
        )
```

r-b-g-b (approved):
One minor tweak, then good to merge -- great work!

hooks/pre_prompt.py
pjbull: Would be good to tell people how to upgrade (e.g., `pip install -U cookiecutter-data-science`)

r-b-g-b: Good call -- updated suggestion.

hooks/pre_prompt.py
```diff
@@ -3,10 +3,11 @@
...
 if __name__ == "__main__":
-    if version <= "2.0.0":
+    if version < "2.0.1":
         warnings.warn(
             "It looks like you're using a version of CCDS that "
-            "always uses the latest template. For more stable behavior, "
-            "update to a newer version of CCDS.",
+            "defaults to using the latest template. For more stable behavior, "
+            "update to CCDS version 2.0.1 or later. To upgrade to the latest "
+            "version, run 'pip install -U cookiecutter-data-science'",
```
pjbull, line 11: I might say "Upgrade to the latest with your package manager, for example `pip install -U cookiecutter-data-science`."

pjbull (approved):

PR #349, opened 2024-03-18
Minor CSS Tweaks

chrisjkuch:
Fixes a few issues where render-deployed version looks different from local.

Specifically, this PR fixes:

* Border color on infobox
* Sizing of code blocks (min-width 35rem, but flexes to grow)
  * Makes sizing of multi-tabbed menu same as underlying code box

### DIFF
--- a/docs/docs/css/extra.css
+++ b/docs/docs/css/extra.css
@@ -35,10 +35,20 @@
     word-break: break-word;
 }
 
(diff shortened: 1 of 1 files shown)


pjbull:
Thanks @chrisjkuch! Would you mind also changing this "note" box I introduced to an info box with the theme colors?

image

PR #350, opened 2024-03-19
Add commandline options table

pjbull:
Adds page with all commandline options and documentation for them to the docs.

Generated from JSON so in the future we can use it for CLI man pages, etc.


pjbull:
C'mon render. SMDH.
```
  File "/opt/render/project/src/docs/scripts/configuration-table.py", line 22
    if choices := item.get("choices", None):
                ^
SyntaxError: invalid syntax
==> Build failed 😞
```

pjbull:
new page here: https://cookiecutter-data-science-pr-350.onrender.com/all-options/

chrisjkuch:
> C'mon render.

I'm still totally flummoxed as to why my render deploys look different from my local.

chrisjkuch (commented):
I really like the pattern of auto-generated docs.

I find the table a little hard to read and wonder whether there might be better ways of visualizing the information hierarchy here, like reducing the table to just one or two columns and treating the info for each prompt more like a card element. I'm happy to take a stab at this before or after we merge this in.

jayqi:
I agree that this table feels hard to read. Might consider a list, or sublists within table cells by combining columns (I think this may be along the lines of @chrisjkuch's suggestion).

pjbull:
Yeah, I'm definitely open to other formats. I like the density of a table, but because of the levels in the tree we end up with too much whitespace and columns that are super narrow.

Maybe we just remove the top level into page sections? That way they show up in the nav, and we can lop off 2 columns.

pjbull:
> I'm still totally flummoxed as to why my render deploys look different from my local.

Is this preview better? I think I moved us from 3.7 -> 3.11 on the builds, which definitely could have been part of it.

chrisjkuch:
image
I think it's better! (gotta go fast)

pjbull:
Updated to use sections for the top-level instead:
https://cookiecutter-data-science-pr-350.onrender.com/all-options/

docs/scripts/configuration-table.py
```diff
@@ -0,0 +1,172 @@
...
+        ),
+        _table_row(["---"] * 4),
+    ]
+
+
+def _new_section(item, item_type, default, description, more_info=""):
+    return [
+        f"## {item.replace('_', ' ').title()}",
+        "",
+        f"**Type:** {item_type}",
```
chrisjkuch, line 30: My only potential build here is to also include the literal name of the `item` as a `Prompt: {item}` field, but that may be redundant / unnecessary.

chrisjkuch (approved):
Looks better & good to me!

jayqi (approved):

PR #359, opened 2024-05-01
converted to pyproject.toml and made project library-aware with namespace

stellarshenson:
The following changes were made:
- dropped setup.py in favour of pyproject.toml
- added clean and build targets to makefile
- made src library-centric with namespace being the src module root


stellarshenson:
fixes required and also I didn't check the v2 branch

PR #362, opened 2024-05-15
Updates from review of V2 docs

pjbull:
Per the comments, does the following:

 - Lists packages for "basic" option (needed some refactoring to get those at docs render time)
 - Have headers same width as body text
 - Add link to announcement blog post at top of docs
 - Update `tree` output at end of index page
 - Update opinion pages formatting to be less varied
 - Add DD footer to nav links
 - Sets min Python version to 3.8
 - Add miniconda, which was removed from MacOS images (see https://github.com/actions/runner-images/issues/9262 )

### DIFF
--- a/docs/docs/opinions.md
+++ b/docs/docs/opinions.md
@@ -2,9 +2,9 @@
 
 The default project structure reflects certain opinions about how to do collaborative data science work. These opinions grew out of our own experiences with what works and what doesn't. Some of these opinions are about workflows, and others are about tools that can make the process easier. These opinions are discussed below. If you have any thoughts, please [contribute or share them](contributing.md).
 
-## Data analysis is a directed acyclic graph
+### Data analysis is a directed acyclic graph
 
-> Don't _ever_ edit your raw data, especially not manually, and _especially_ not in Excel.
+_Don't ever edit your raw data. Especially not manually. And especially not in Excel._
 
 The most important features of a quality data analysis are **correctness** and **reproducibility**—anyone should be able to re-run your analysis using only your code and raw data and produce the same final products. The best way to ensure correctness is to test your analysis code. **The best way to ensure reproducibility is to treat your data analysis pipeline as a directed acyclic graph ([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))**. This means each step of your analysis is a node in a directed graph with no loops. You can run through the graph forwards to recreate any analysis output, or you can trace backwards from an output to examine the combination of code and data that created it.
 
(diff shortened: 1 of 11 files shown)


docs/docs/opinions.md
```diff
@@ -2,9 +2,9 @@
...
 
-## Data analysis is a directed acyclic graph
+### Data analysis is a directed acyclic graph
 
-> Don't _ever_ edit your raw data, especially not manually, and _especially_ not in Excel.
+_Don't ever edit your raw data. Especially not manually. And especially not in Excel._
 
 The most important features of a quality data analysis are **correctness** and **reproducibility**—anyone should be able to re-run your analysis using only your code and raw data and produce the same final products. The best way to ensure correctness is to test your analysis code. **The best way to ensure reproducibility is to treat your data analysis pipeline as a directed acyclic graph ([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))**. This means each step of your analysis is a node in a directed graph with no loops. You can run through the graph forwards to recreate any analysis output, or you can trace backwards from an output to examine the combination of code and data that created it.
```
chrisjkuch, line 9:
```suggestion
The most important features of a quality data analysis are correctness and reproducibility—anyone should be able to re-run your analysis using only your code and raw data and produce the same final products. The best way to ensure correctness is to test your analysis code. **The best way to ensure reproducibility is to treat your data analysis pipeline as a directed acyclic graph ([DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph))**. This means each step of your analysis is a node in a directed graph with no loops. You can run through the graph forwards to recreate any analysis output, or you can trace backwards from an output to examine the combination of code and data that created it.
```
Responding here to feedback that there's too much bold text (IMO this feedback is quite right)

docs/docs/opinions.md
```diff
@@ -2,11 +2,11 @@
 
 The default project structure reflects certain opinions about how to do collaborative data science work. These opinions grew out of our own experiences with what works and what doesn't. Some of these opinions are about workflows, and others are about tools that can make the process easier. These opinions are discussed below. If you have any thoughts, please [contribute or share them](contributing.md).
 
-## Data analysis is a directed acyclic graph
+### Data analysis is a directed acyclic graph
 
-> Don't _ever_ edit your raw data, especially not manually, and _especially_ not in Excel.
+_Don't ever edit your raw data. Especially not manually. And especially not in Excel._
```
chrisjkuch, line 7: Was the un-quote blocking here intentional? I think it's probably fine to keep this as-is, but it breaks the format we use everywhere else of quote-blocking the subtitle.

chrisjkuch (approved):
one small formatting note, otherwise LGTM!

docs/docs/opinions.md
pjbull: Yeah, it was. This page in particular got feedback that the styles were too busy at first glance.

PR #372, opened 2024-05-30
Remove config import from init to fix scaffold usage

### LINKED ISSUE #370: config import fails when using V2 scaffolding
At:
https://github.com/drivendataorg/cookiecutter-data-science/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/%7B%7B%20cookiecutter.module_name%20%7D%7D/__init__.py#L1

Should be `from {{ cookiecutter.module_name }} import config`  # noqa: F401

------------

### Discussed in https://github.com/drivendataorg/cookiecutter-data-science/discussions/369

<div type='discussions-op-text'>

<sup>Originally posted by **wrighteap** May 27, 2024</sup>
Hi!

New cookiecutter user here. I just tried using your new v2 release.

I created a project project, repo repo and module module, using the code scaffold option. I selected conda as my environment manager, and requirements.txt for my dependency file. 

pip list shows the package module installed at /path/to/repo. When I try to import module, I get a ModuleNotFoundError: No module named 'config', due to the import config under module/__init__.py -- Is this a design choice to invite me to add /path/to/repo/module to sys path or to modify the config directly?

Sorry in advance if the question is inane or lacking in detail - happy to be told so or asked for additional information.</div>

pjbull:
- Remove config import from `__init__.py` (not needed and also breaks since it is there in non-scaffold scenario)
 - Refactor some repeated code
 - Add import and import config tests
 - Add changelog

Closes #370

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,9 @@
+# cookiecutter-data-science Changelog
+
+## UNRELEASED
+
+- Fixes issue with scaffold code that import of config did not work. Adds testing of imports to test suite. (Issue [#370](https://github.com/drivendataorg/cookiecutter-data-science/issues/370))
+
+## v2.0.0 (2024-05-22)
+
+- Released version 2.0.0! :tada: See [docs](https://cookiecutter-data-science.drivendata.org/) and [announcement blog post](https://drivendata.co/blog/ccds-v2) for more information.
--- a/ccds/__main__.py
+++ b/ccds/__main__.py
@@ -28,9 +28,9 @@ def default_ccds_main(f):
     """Set the default for the cookiecutter template argument to the CCDS template."""
 
     def _main(*args, **kwargs):
-        f.params[
-            1
-        ].default = "https://github.com/drivendataorg/cookiecutter-data-science"
+        f.params[1].default = (
+            "https://github.com/drivendataorg/cookiecutter-data-science"
+        )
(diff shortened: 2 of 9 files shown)


jayqi:
> not needed

The main thing that happens in `config` right now that is valuable to auto-execute is this bit:

https://github.com/drivendataorg/cookiecutter-data-science/blob/bbfa1511ff724731ddd482bcffcd62fb1e6b1b83/%7B%7B%20cookiecutter.repo_name%20%7D%7D/%7B%7B%20cookiecutter.module_name%20%7D%7D/config.py#L24-L32

pjbull:
> The main thing that happens in config right now that is valuable to auto-execute is this bit:

Ah, that's a good point. Added back in.

jayqi (approved):

chrisjkuch (approved):
LGTM!

Just summarizing for my own understanding / posterity:

* We weren't appropriately removing `import config` from `__init__.py` when we removed all the boilerplate code, which failed at import time because `config.py` no longer existed
* When the boilerplate was present, we weren't importing `config` from the right place leading to an import error
* We now import the config from the right place and remove it when boilerplate is removed, and we test that both of those work as expected in the live tests

pjbull:
> Just summarizing for my own understanding / posterity:

Nailed it! 🔨

PR #387, opened 2024-06-19
Add ruff as default linting + formatting option

### LINKED ISSUE #388: Ruff lint configuration names are incorrect
To reproduce:

1. Create a new cookiecutter-data-science repo
2. Run `ruff lint .`

Error:

```
Ruff: Lint failed (ruff failed
  Cause: Failed to parse /home/.../diagnostics/pyproject.toml
  Cause: TOML parse error at line 30, column 12
   |
30 | [tool.ruff.lint.isort]
   |            ^^^^
unknown field `known_first_party`, expected one of `force-wrap-aliases`, `force-single-line`, `single-line-exclusions`, `combine-as-imports`, `split-on-trailing-comma`, `order-by-type`, `force-sort-within-sections`, `case-sensitive`, `force-to-top`, `known-first-party`, `known-third-party`, `known-local-folder`, `extra-standard-library`, `relative-imports-order`, `required-imports`, `classes`, `constants`, `variables`, `no-lines-before`, `lines-after-imports`, `lines-between-types`, `forced-separate`, `section-order`, `default-section`, `no-sections`, `detect-same-package`, `from-first`, `length-sort`, `length-sort-straight`, `sections`
)
```

It appears that the `pyproject.toml` ruff configuration setting names are incorrect -- they should use hyphens rather than underscores as separators.

https://docs.astral.sh/ruff/configuration/

chrisjkuch:
Closes #374
Closes #388 

This PR adds [ruff](https://docs.astral.sh/ruff/) as the default linting and formatting option for new ccds projects.

## Implementation notes

* I've slightly altered the Makefile command tests so that they now run within each venv harness. I check to ensure that linting and formatting run without error and that nothing is reformatted.
  * For whatever reason, I found that ruff piped to stdout and flake8/black/isort piped to stderr. Maybe there's a nuance here or I've done something wrong?
* I've cleaned up the generated `pyproject.toml` and removed `setup.cfg` if using ruff since it's no longer needed. 
* I didn't stray very far from the default ruff config, changing only the line length. If we are opinionated about the default set of linting / formatting options, happy to add those, but I have no strong preferences.

## Discussion

As mentioned, this PR makes ruff the default option. I think there's good arguments for this that mostly boil down to "it's simpler + faster":

* one tool for linting + formatting so fewer dependencies and simpler config
* no `setup.cfg` file so everything can live in `pyproject.toml`
* ruff is [quite fast](https://docs.astral.sh/ruff/)

However, ruff is still (relatively) new, whereas flake8/black/isort is more mature, and most projects won't notice the speed difference until they get quite large. I could be persuaded that we should introduce ruff without making it the default option.

### DIFF
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -35,6 +35,7 @@ clean:
 ## Lint using ruff (use `make format` to do formatting)
 .PHONY: lint
 lint:
+	ruff format --check
 	ruff check
 
 ## Format source code with ruff
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -35,13 +35,13 @@ clean:
 ## Lint using ruff (use `make format` to do formatting)
 .PHONY: lint
 lint:
-	ruff check {{ cookiecutter.module_name }}
+	ruff check
 
 ## Format source code with ruff
 .PHONY: format
 format:
-	ruff check --fix {{ cookiecutter.module_name }}
-	ruff format {{ cookiecutter.module_name }}
+	ruff check --fix
+	ruff format
 {% elif cookiecutter.linting_and_formatting == 'flake8+black+isort' %}
 ## Lint using flake8, black, and isort (use `make format` to do formatting)
 .PHONY: lint
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -40,21 +40,22 @@ lint:
 ## Format source code with ruff
 .PHONY: format
 format:
-	ruff check --select I --fix {{ cookiecutter.module_name }}
+	ruff check --fix {{ cookiecutter.module_name }}
 	ruff format {{ cookiecutter.module_name }}
 {% elif cookiecutter.linting_and_formatting == 'flake8+black+isort' %}
 ## Lint using flake8, black, and isort (use `make format` to do formatting)
 .PHONY: lint
 lint:
 	flake8 {{ cookiecutter.module_name }}
-	isort --check --diff --profile black {{ cookiecutter.module_name }}
-	black --check --config pyproject.toml {{ cookiecutter.module_name }}
+	isort --check --diff {{ cookiecutter.module_name }}
+	black --check --verbose {{ cookiecutter.module_name }}
 
 ## Format source code with black
 .PHONY: format
 format:
-	isort --profile black {{ cookiecutter.module_name }}
-	black --config pyproject.toml {{ cookiecutter.module_name }}
+	isort {{ cookiecutter.module_name }}
+	black --verbose {{ cookiecutter.module_name }}
+	exit -1
 {% endif %}
 
 {% if not cookiecutter.dataset_storage.none %}
--- a/{{ cookiecutter.repo_name }}/Makefile
+++ b/{{ cookiecutter.repo_name }}/Makefile
@@ -48,14 +48,13 @@ format:
 lint:
 	flake8 {{ cookiecutter.module_name }}
 	isort --check --diff {{ cookiecutter.module_name }}
-	black --check --verbose {{ cookiecutter.module_name }}
+	black --check {{ cookiecutter.module_name }}
 
 ## Format source code with black
 .PHONY: format
 format:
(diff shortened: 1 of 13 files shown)


{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -31,7 +31,19 @@ clean:
...
+{% if cookiecutter.linting_and_formatting == 'ruff' %}
+## Lint using ruff (use `make format` to do formatting)
+.PHONY: lint
+lint:
+	ruff check {{ cookiecutter.module_name }}
+
+## Format source code with ruff
+.PHONY: format
+format:
+	ruff check --select I --fix {{ cookiecutter.module_name }}
```
jayqi, line 43: I don't like `--select I` here. We should use the `fixable` or `unfixable` configuration in `pyproject.toml`.

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -41,7 +53,9 @@ lint:
 ## Format source code with black
 .PHONY: format
 format:
+	isort --profile black {{ cookiecutter.module_name }}
```
jayqi, line 56: We should move `--profile black` to `pyproject.toml` instead of having it here, so that configuration is all in one place.

jayqi (changes requested):

{{ cookiecutter.repo_name }}/Makefile
chrisjkuch: Agreed on removing --select. Do you have an opinion on whether we should keep the default (everything is fixable) or specify fixable vs unfixable categories?

jayqi: In my other projects so far, I've been setting `F` as unfixable and it's been working for me so far. I haven't looked that hard at the rulesets though. 

https://github.com/drivendataorg/erdantic/blob/664a3f71e22225635bc95e26578312c9b19dd363/pyproject.toml#L54-L60

jayqi:
I'm in favor of having the default to be Ruff. It's clear that Ruff is the emerging standard, and they describe themselves as production-ready.

There are major projects that are fully using Ruff for linting and formatting. 

- [Pandas](https://github.com/pandas-dev/pandas/blob/c46fb76afaf98153b9eef97fc9bbe9077229e7cd/pyproject.toml#L181-L188)
- [pytest](https://github.com/pytest-dev/pytest/blob/57bc6df510899ba77de5129d975f148c4a868725/.pre-commit-config.yaml#L4-L7)

Lots of other major projects (e.g., pip, matplotlib, PyTorch) are using Ruff for linting over flake8, though are still using black for now.

szapp:
It seems that slight adjustments to the linter rules are necessary to ensure similar behavior as the flake8/black/isort alternative. Most of the rules that flake8 ignores in the `setup.cfg` are ignored in the [defaults of ruff](https://docs.astral.sh/ruff/configuration) as well, except for `E731` if I am not mistaken.

If I am not wrong, these additional lines in `pyproject.toml` should do.

```toml
[tool.ruff.lint]
ignore = [
  "E731"  # lambda-assignment: Do not assign a lambda expression, use a def
]
```

{{ cookiecutter.repo_name }}/Makefile
chrisjkuch: I'm going to leave the selection at the default, with the exception of adding `"I"` to get isorting.

chrisjkuch:
@szapp thanks for catching the isort fix! I've opted to keep the default selections in ruff for now - I don't know that we're necessarily aiming for exact parity with black, as opposed to just a good starting point for a project.

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -31,17 +31,31 @@ clean:
 	find . -type f -name "*.py[co]" -delete
 	find . -type d -name "__pycache__" -delete
 
-## Lint using flake8 and black (use `make format` to do formatting)
+{% if cookiecutter.linting_and_formatting == 'ruff' %}
+## Lint using ruff (use `make format` to do formatting)
+.PHONY: lint
+lint:
+	ruff check
```
jayqi, line 38:
```suggestion
	ruff format --check
	ruff check
```

jayqi (changes requested):

jayqi (approved):

PR #392, opened 2024-07-04
fix: use makefile global variable for python interpreter in help command

philipp91roberto:
The help makefile command is not using the global variable that defines the python interpreter set. Hence, even setting the variable for example to "python3", the help command throws an error when run.


pjbull:
Thanks @philipp91roberto, this seems reasonable to me! Will merge on passing tests.

PR #397, opened 2024-08-01
Update directory tree in README template

maoding:
This PR fixes the directory tree in the README template.

It seems that the layout structure has changed with cookiecutter-data-science v2 thus the generated structure does not match the directory tree shown in the README.

I copy pasted the directory tree from the top-level README and pasted it into the template.


pjbull:
Thanks @maoding!

PR #407, opened 2024-12-28
Added options to cookiecutter cli for version control, no changes mad…

GatlenCulp:
…e yet to rendering

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,13 @@ manual_test/
 
 # other local dev info
 .vscode/
+.history/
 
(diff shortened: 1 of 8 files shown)


jayqi:
I'd be in favor of initializing a local Git repository, whether it's an option or the default option. IMO the discussion in #380 is outdated now, and in 2025, Git is de facto the one standard VCS software. 

I'm not as keen on the options to push to GitHub. Writing to an external system automatically at the end of creating a stub project feels heavy-handed.

PR #408, opened 2024-12-29
Added support for the uv package and project manager

GatlenCulp:
### Summary
This PR integrates [uv](https://docs.astral.sh/uv/) — a fast Python package and project manager written in Rust by the team that made the ruff linter — as an environment manager. Updates made to the Makefile and CLI to support this.

GatScreen_2024-12-30_16 00 24@2x

### Key Changes
- **Makefile**  
  - `make create_environment` - Added a target to creates a virtual environment with uv and installs the requested python version using uv
  - `make requirements` simply installs dependencies with `uv pip install -r requirements.txt` which is faster than normal pip. This does not generate a lockfile and the dependencies will be removed if `uv sync` is run. (see below for more info)

### Testing
- `uv` added as an additional environment manager in `test_creation.py` and a harness was added in similar style to the other environment managers. Tests passing locally.  
- GitHub actions tests are passing [passing](https://github.com/GatlenCulp/gatlens-opinionated-template/actions/runs/12552502703)

### Recommendations
- Consider adding an option with **no** dependency file and [track dependencies in `pyproject.toml` only](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/). This allows for using the higher-level and convenient `uv sync` or `uv add` commands which generate a lockfile. This is also becoming more standard.
	- Using `uv add -r requirements.txt` after removing `-e .` from the file will add the dependencies to `pyproject.toml` and generate the lockfile. This might be a better implementation for `make requirements`


{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -15,8 +15,13 @@ PYTHON_INTERPRETER = python
 .PHONY: requirements
 requirements:
 	{% if "requirements.txt" == cookiecutter.dependency_file -%}
+	{% if "uv" == cookiecutter.environment_manager -%}
+	uv pip install pip
```
jayqi, line 19: I don't think we need to install pip if uv is being chosen as the environment manager.

pyproject.toml
```diff
@@ -42,3 +42,7 @@ ccds = "ccds.__main__:main"
 "Source Code" = "https://github.com/drivendataorg/cookiecutter-data-science/"
 "Bug Tracker" = "https://github.com/drivendataorg/cookiecutter-data-science/issues"
 "DrivenData" = "https://drivendata.co"
+
+[tool.pytest.ini_options]
+testpaths = "./tests"
+addopts = "-vv --color=yes"
```
jayqi, line 48: As a general code style thing, can you make sure all files end in a newline?

tests/virtualenv_harness.sh
```diff
@@ -56,5 +56,4 @@ fi
 
 make requirements
 
-run_tests $PROJECT_NAME $MODULE_NAME
-
+run_tests $PROJECT_NAME $MODULE_NAME
```
jayqi, line 59: Per other comment about having files end in a newline, this file shouldn't have any changes.

jayqi (changes requested):

tests/virtualenv_harness.sh
chrisjkuch: @jayqi the file on master has two newlines at the end of it, so technically this change is making the file correctly formatted. Can move this change to a separate formatting PR though if desired.

jayqi (approved):
Looks good to me.

As a note: When we write the changelog entry for this, we should say that we've added support for the `uv pip` interface. 

We don't support the uv project interface yet, which @GatlenCulp mentioned in the PR description. This requires support for pyproject.toml (new issue #431). We may want to use PEP 735 dependency groups. It'll make sense to work on that in the near future—even regular Pip will be releasing support for them in their April release.

PR #409, opened 2024-12-30
Add PyTest IDs for better readability and filtering

GatlenCulp:
This small PR adds dynamically generated IDs for the config parameters. This makes it immediately clear which configuration failed and [allows for filtering with `pytest -k EXPRESSION`](https://docs.pytest.org/en/6.2.x/usage.html#specifying-tests-selecting-tests). Ex: To only run the tests using conda and mkdocs you can do `pytest -k "conda and mkdocs"`.

```python
def pytest_generate_tests(metafunc):
    # setup config fixture to get all of the results from config_generator
    def make_test_id(config): # NEW
        return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"
    
    if "config" in metafunc.fixturenames:
        metafunc.parametrize(
            "config",
            config_generator(metafunc.config.getoption("fast")),
            ids=make_test_id # NEW
        )
```

Before:
GatScreen_2024-12-30_17 39 15@2x

After:
GatScreen_2024-12-30_17 44 33@2x

(Screenshots from VSCode Test Explorer)

_The changes to `pyproject.toml` and `.gitignore` are small changes for my local development which you're free to add or ignore._

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -108,6 +108,7 @@ def pytest_generate_tests(metafunc):
     # setup config fixture to get all of the results from config_generator
     def make_test_id(config):
         return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"
+
     if "config" in metafunc.fixturenames:
         metafunc.parametrize(
             "config",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -113,7 +113,7 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize(
             "config",
             config_generator(metafunc.config.getoption("fast")),
-            ids=make_test_id
+            ids=make_test_id,
         )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -108,7 +108,6 @@ def pytest_generate_tests(metafunc):
     # setup config fixture to get all of the results from config_generator
     def make_test_id(config):
         return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"
-    
     if "config" in metafunc.fixturenames:
         metafunc.parametrize(
             "config",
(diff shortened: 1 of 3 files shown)


chrisjkuch:
Sorry for the delay on this @GatlenCulp - would you mind fixing the linting issue so I can run the tests?

tests/conftest.py
```diff
@@ -106,9 +106,14 @@ def fast(request):
 
 def pytest_generate_tests(metafunc):
     # setup config fixture to get all of the results from config_generator
+    def make_test_id(config):
+        return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"
+
```
chrisjkuch, line 111:
```suggestion

```

tests/conftest.py
```diff
@@ -106,9 +106,13 @@ def fast(request):
 
 def pytest_generate_tests(metafunc):
     # setup config fixture to get all of the results from config_generator
+    def make_test_id(config):
+        return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"
```
chrisjkuch, line 110:
```suggestion
        return f"{config['environment_manager']}-{config['dependency_file']}-{config['docs']}"

```

tests/conftest.py
```diff
@@ -106,9 +106,14 @@ def fast(request):
...
```
chrisjkuch, line 110: I think this would be more informative if we named the config with whether it contained the basic pydata packages rather than docs, since the former has more impact than the latter.
```suggestion
        return f"{config['environment_manager']}-{config['dependency_file']}-{config['pydata_packages']}"
```

chrisjkuch (changes requested):
One small comment, then looking for another reviewer and we should be good to go!

chrisjkuch (approved):
Thanks again for the PR - I've made a few slight tweaks and now looks good to me! Will get another reviewer to check my work, then we're planning to merge for a 2.0.2 release with #387 and #408

jayqi (approved):
LGTM

PR #424, opened 2025-02-17
Deprecate Python 3.8 support, add 3.13 support and tests

### LINKED ISSUE #423: Update CI to supported Python versions
As of writing, this is 3.9 to 3.13.

https://devguide.python.org/versions/

chrisjkuch:
Closes #423

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: "3.11"
           cache: "pip"
           cache-dependency-path: |
             pyproject.toml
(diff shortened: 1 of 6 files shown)


pjbull:
@chrisjkuch We want to grab the CI workflows too:
https://github.com/search?q=repo%3Adrivendataorg%2Fcookiecutter-data-science%20%223.8%22&type=code

Also, not sure we need to update the v1 docs since those are just for posterity and we don't test/support on it.

.github/workflows/tests.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.8, 3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.9, "3.10", "3.11", "3.12", "3.13"]
```
chrisjkuch, line 52: I think grabs the CI workflows - am I missing something?

pjbull (approved):
LGTM!

Issue #416, opened 2025-01-23
Env name does not match folder name if folder has `-` in it


pjbull:
Repro steps:
 - Create new project with a `-` in the folder name, e.g. `project-name`
 - Run `make_create_environment`

Result:
Created environment has `_` instead, so it does not match the project name (e.g., `project_name` instead of `project_name`).

Env manager: conda

chrisjkuch:
@pjbull I'm not able to reproduce this - here's what I tried:

Conda env with `requirements.txt`:
```
❯ make manual-test
...
project_name (project_name): project-name
repo_name (project-name): repo-name
module_name (project_name): module-name
...
Select environment_manager
    1 - virtualenv
    2 - conda
    3 - pipenv
    4 - none
    Choose from [1/2/3/4] (1): 2
Select dependency_file
    1 - requirements.txt
    2 - environment.yml
    3 - Pipfile
    Choose from [1/2/3] (1): 1
...

❯ cd manual_test/repo-name
❯ make create_environment
conda create --name repo-name python=3.10 -y  # Environment named correctly
```

With `environment.yml`:
```
❯ make manual-test
...
project_name (project_name): project-name
repo_name (project-name): repo-name
module_name (project_name): module-name
...
Select environment_manager
    1 - virtualenv
    2 - conda
    3 - pipenv
    4 - none
    Choose from [1/2/3/4] (1): 2
Select dependency_file
    1 - requirements.txt
    2 - environment.yml
    3 - Pipfile
    Choose from [1/2/3] (1): 2
...

❯ cd manual_test/repo-name
❯ make create_environment
conda env create --name repo-name -f environment.yml  # Environment named correctly
...
❯ head environment.yml  # Environment named correctly
name: repo-name
...

We do replace spaces with underscores in the repo name, and hyphens with underscores in the module name: https://github.com/drivendataorg/cookiecutter-data-science/blob/fb791fa2dcf52159d19194162b45e872deda14be/ccds.json#L2-L4

jayqi:
I am also unable to reproduce this problem.

It looks like the generated Makefile uses `{{ cookiecutter.repo_name }}` as the environment name everywhere, and running through project generation, my default value for `repo_name` retains all hyphens as expected. 

https://github.com/drivendataorg/cookiecutter-data-science/blob/fb791fa2dcf52159d19194162b45e872deda14be/%7B%7B%20cookiecutter.repo_name%20%7D%7D/Makefile#L5

PR #426, opened 2025-02-17
Fix pre-prompt ccds version error

### LINKED ISSUE #425: No module named 'ccds.version' Pre-Prompt Hook script failed
Not sure what happened that broke but now anytime I try to initiate a new project with `ccds` I receive the following error

```
from ccds.version import __version__ as version
ModuleNotFoundError: No module named 'ccds.version'
Pre-Prompt Hook script failed
```

chrisjkuch:
Closes #425 

This fixes an issue introduced in #336 that added a pre-prompt hook to warn about the newly introduced versioning system.

However, the pre-prompt script was unable to access module-level imports from the ccds package (I'm still trying to figure out / think about why that's the case).

Migrating version extraction from a separate `version.py` module into `__init__.py` seems to fix this, since the version can be determined directly by importlib without needing separate imports from the ccds module.


hooks/pre_prompt.py
```diff
@@ -1,9 +1,9 @@
 import warnings
 
-from ccds.version import __version__ as version
+from ccds import __version__
 
 if __name__ == "__main__":
-    if version < "2.0.1":
+    if __version__ < "2.0.1":
```
jayqi, line 6: Is there a scenario where this would still break? You could potentially put this in a try-except.

chrisjkuch: Yeah I think that's a good idea regardless.

pjbull (approved):
Change looks reasonable. How did you manually test different configs?

hooks/pre_prompt.py
```diff
@@ -0,0 +1,16 @@
+import warnings
+
+try:
+    from ccds import __version__
+except ImportError:
+    __version__ = None  # ccds is not installed
```
r-b-g-b, line 6:
```suggestion
    importlib.metadata import version, PackageNotFoundError
    try:
        __version__ = version("cookiecutter-data-science")  # ccds is not installed
    except importlib.metadata.PackageNotFoundError:
        __version__ = "1.0.0"  # or None
```

I wanted to see if I could get this warning to show up so I:

- Install `ccds` `pip install cookiecutter-data-science==2.0.0`
- Create new project using the template from this branch `ccds -c 425-version-bugfix`
- No warning shows up!

In hindsight that makes sense, bc:
- `ccds.__version__` is not defined prior to this commit (`from ccds import __version__` will fail with `ImportError` on `cookiecutter-data-science==2.0.0`)
- In the hook, the `except` block sets `__version__ = None`
- `if __version__ is not None` prevents warning from displaying

But we want this warning to show up if someone is using `2.0.0`. One way to do that would be to use `importlib.metdata.version` in the hook itself (in the suggestion). Alternatively, we can remove the `if __version__ is not None` condition to display the warning whenever the `ccds.__version__` does not exist (all ccds versions prior to `2.0.1`).

chrisjkuch: Thank you for catching this @r-b-g-b! When I was testing locally, it seemed like it was working - then I realized that I had an environment issue (combo of conda base env + pipx) that made it look as though it was doing the right thing when it was not.

I've added a fix that uses more specific import errors to suss out whether ccds is not installed (throws a `ModuleNotFoundError`) and whether it is installed but an older version (throws an `ImportError) to get the desired behavior. Can you confirm that this new fix works for you? I've uninstalled all the ccds libraries on my machine and progressively added things in the following way:

```bash
pip uninstall ccds
pipx uninstall ccds

git clone -b 425-version-bugfix git@github.com:drivendataorg/cookiecutter-data-science.git test-ccds
cd test-ccds
uv venv --python 3.10
uv pip install cookiecutter

# should *not* print the 2.0.1 warning, ccds not installed. will print cookiecutter deprecation message
cookiecutter . 
uv pip install cookiecutter-data-science

# *should* print the 2.0.1 warning, ccds 2.0.0 installed
uv run ccds .  
sed -i -e 's/2\.0\.0/2.0.1/' pyproject.toml
uv pip install -e .

# should *not* print the error, 2.0.1 installed
uv run ccds .  
```

r-b-g-b: Heyyyy it works! I like the way you implemented it with `ModuleNotFoundError` / `ImportError`.

r-b-g-b (approved):
:tada:

jayqi:
@chrisjkuch is this good to go?

jayqi:
@chrisjkuch is #416 really a blocker? It seems like (1) neither you nor I can reproduce it, and (2) if is is actually a problem, an unrelated specific bug that we could fix separately.

chrisjkuch:
Agreed, we can push #416 / handle separately. If @pjbull is onboard, we can merge and then update to / release 2.0.1!

pjbull:
Yeah, let's do it! ✨

PR #436, opened 2025-03-07
Add pyproject.toml as a dependencies file option

### LINKED ISSUE #431: Add pyproject.toml as dependency file option
`pyproject.toml` files have two standard ways to support specifying dependencies—the `[project].dependencies` field and `[dependency-groups]`—which can obviate the need for a separate `requirements.txt` file. 

We have a design decision about where to add dependencies.

1. Add everything together into one field (the same way we currently do `requirements.txt`)
2. Split up package runtime dependencies (under the `dependencies` field) and development dependencies (using dependency groups)

My preference is Option 2, as its follows the conceptual design of `pyproject.toml` more faithfully. It also would simplify the transition of the project to a published package. However, it comes with the tradeoff of introducing some added complexity to the user of having two lists of dependencies (though in a single config file).

If we choose to use dependency groups at all (e.g., choosing Option 2, or Option 1 but using a dependency group), then this is something that uv as an environment manager supports already. Pip has [merged support for it](https://github.com/pypa/pip/pull/13065) but it hasn't been released yet. It looks like it'll be part of the [25.1 release](https://github.com/pypa/pip/milestone/78) coming in April.

jayqi:
Adds `pyproject.toml` as a dependency file option. Closes #431.


pyproject.toml
```diff
@@ -32,6 +32,7 @@ requires-python = ">=3.9"
 dependencies = [
   "click",
   "cookiecutter",
+  "tomlkit",
```
jayqi, line 35: TOML Kit is a zero-dependency TOML-writing library published by the Poetry organization. Having a structured and less hacky way of adding chosen dependencies `pyproject.toml` file feels worth it to me.

tests/test_creation.py
```diff
@@ -146,7 +146,7 @@ def verify_files(root, config):
 
     existing_files = [f.relative_to(root) for f in root.glob("**/*") if f.is_file()]
 
-    assert sorted(existing_files) == sorted(expected_files)
+    assert sorted(existing_files) == sorted(set(expected_files))
```
jayqi, line 149: `pyproject.toml` gets added twice, since it's already there even if you don't choose it as a dependency file.

{{ cookiecutter.repo_name }}/Makefile
```diff
@@ -21,6 +21,12 @@ requirements:
 	$(PYTHON_INTERPRETER) -m pip install -U pip
 	$(PYTHON_INTERPRETER) -m pip install -r requirements.txt
 	{% endif -%}
+	{% elif "pyproject.toml" == cookiecutter.dependency_file -%}
+	{% if "uv" == cookiecutter.environment_manager -%}
+	uv sync
```
jayqi, line 26: `uv sync` is nicer than `uv pip install -e .`. You get a lockfile, and it makes sure the environment matches the declared dependencies.

pjbull (approved):
Looks great to me! ✨

ccds/hook_utils/dependencies.py
```diff
@@ -40,6 +42,15 @@ def write_dependencies(
...
 
+    elif dependencies == "pyproject.toml":
+        with open(dependencies, "r") as f:
+            doc = tomlkit.parse(f.read())
+        doc["project"].add("dependencies", sorted(packages))
+        doc["project"]["dependencies"].multiline(True)
+
+        with open(dependencies, "w") as f:
+            f.write(tomlkit.dumps(doc))
+
```
chrisjkuch, line 53: Only thing worth noting here is that folks using 2.1.0 with `-c master` will get a silent failure populating the pyproject.toml with dependencies. This code block won't exist for them and they won't get a warning about upgrading to 2.2.

PR #439, opened 2025-03-10
Add back isort configuration that was removed in #387

### LINKED ISSUE #388: Ruff lint configuration names are incorrect

jayqi:
#387 changed our default behavior by removing configuration. This PR adds them back in. It also fixes #388 by using the correct names.

### DIFF
--- a/{{ cookiecutter.repo_name }}/pyproject.toml
+++ b/{{ cookiecutter.repo_name }}/pyproject.toml
@@ -29,6 +29,8 @@ exclude = '''
 
 [tool.isort]
 profile = "black"
+known_first_party = ["{{ cookiecutter.module_name }}"]
+force_sort_within_sections = true
 {% endif %}
 {% if cookiecutter.linting_and_formatting == 'ruff' %}
 [tool.ruff]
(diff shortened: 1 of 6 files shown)


jayqi:
@pjbull just confirming that this is what you were complaining about on Friday, right?

You prefer

```python
from loguru import logger
from tqdm import tqdm
import typer
```

over

```python
import typer
from loguru import logger
from tqdm import tqdm
```

chrisjkuch (approved):
LGTM

PR #446, opened 2025-03-19
Resolve python version to work with compatible operator

### LINKED ISSUE #444: uv dependency management doesn't use specified Python version for venv
## Reproduce

- Instantiate a template, using `ccds -c master`
- Select:
  -  uv as environment manager
  - pyproject.toml as dependency file
  - Python version **3.10**

- Create a virtual environment

  ```sh
  uv venv

  # Using CPython 3.13.2
  ```

- Run a script

  ```python
  # script.py
  import sys
  print(sys.executable)
  print(sys.version)
  ```

  ```sh
  rm -fr .venv
  uv run script.py

  # Using CPython 3.13.2
  # Creating virtual environment at: .venv
  # Installed 2 packages in 24ms
  # /home/robert/projects/hi/.venv/bin/python
  # 3.13.2 (main, Feb 12 2025, 14:51:17) [Clang 19.1.6 ]
  ```

## Proposed solution

I think the preferred way to specify a project-level Python version is to use the [`.python-version`](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files) file.

r-b-g-b:
Closes #444 

Add a post gen hook to resolves a Python version string to one that behaves as expected with the "[compatible release](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release)" operator which we use in `pyproject.toml` `[requires-python]`, for example:

-   3.13 should result in ~=3.13.0 which installs the latest version of Python that is compatible with 3.13, so the highest patch number for 3.13 (3.13.3 at the time of writing)
-   3.13.2 should result in ==3.13.2 which ensures Python 3.13.2 is used

### DIFF
--- a/ccds/hook_utils/dependencies.py
+++ b/ccds/hook_utils/dependencies.py
@@ -30,14 +30,18 @@ scaffold = [
 ]
 
 
-def resolve_compatible_python_version(python_version):
-    """Resolves a Python version string to one that behaves as expected with the "compatible release"
-    operator.
+def resolve_python_version_specifier(python_version):
+    """Resolves the user-provided Python version string to a version specifier.
 
     Examples:
-    - 3.13 should result in ~=3.13.0 which installs the latest version of Python that is compatible
-      with 3.13, so the highest patch number for 3.13 (3.13.3 at the time of writing)
-    - 3.13.2 should result in ~=3.13.2 which ensures Python 3.13.2 is used
+
+    User provides: 3.12
+    Resolved version specifier: ~=3.12.0
+    Compatible versions: 3.12.0, 3.12.1, 3.12.2, etc.
+
+    User provides: 3.12.2
+    Resolved version specifier: ==3.12.2
+    Compatible versions: 3.12.2
 
     See https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release
     """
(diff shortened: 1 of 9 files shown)


ccds/hook_utils/dependencies.py
```diff
@@ -30,6 +30,43 @@
...
 
 
+def resolve_compatible_python_version(python_version):
+    """Resolves a Python version string to one that behaves as expected with the "compatible release"
+    operator.
+
+    Examples:
+    - 3.13 should result in ~=3.13.0 which installs the latest version of Python that is compatible
+      with 3.13, so the highest patch number for 3.13 (3.13.3 at the time of writing)
+    - 3.13.2 should result in ~=3.13.2 which ensures Python 3.13.2 is used
```
jayqi, line 40: >     - 3.13.2 should result in ~=3.13.2 which ensures Python 3.13.2 is used

This is not correct. `~=3.13.2` is equivalent to `>= 3.13.2, == 3.13.*`, so `3.13.3`, `3.13.4`, etc. will also satisfy it. 

https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

In the case of 3 parts, I think we want to use an exact version match.

jayqi (changes requested):
FYI, I think our CI tests should fail for #444, and I'm working on trying to make that happen in #445.

I think we should merge the two PRs together before merging them into master.

ccds/hook_utils/dependencies.py
r-b-g-b: Aha thanks, I had misunderstood that. Updated, and confirmed that it works when providing <major>.<minor> (use `~=`) and <major>.<minor>.<patch> (use `==`).

jayqi (approved):

PR #451, opened 2025-05-02
Fix terminal animation generation in docs build

chrisjkuch:
When we added testing frameworks to the option set (#447) we did not update the terminal animation inputs so the last one was missing. This made the build fail because we `assert` that all options are present in the generated animation.

To prevent us missing this in the future, I've added building the docs to the `tests` workflow so that we'll fail if the docs do not properly generate.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,12 +5,10 @@ on:
   push:
     branches: [master]
     paths-ignore: 
-      - "docs/**"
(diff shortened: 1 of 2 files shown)


pjbull (approved):
Change LGTM, pending resolving the failures (which look like they are happening on `main` too. Looks like some resolution of dependency problem for the particular combination of packages in one of the configs on Windows.)

PR #460, opened 2025-07-23
Add poetry as an env manager

pjbull:
Adds support for poetry as an env/dependency manager.

 - Add `poetry` to dev requirements for testing
 - Add `poetry` option to dep manager
 - Use `pyproject.toml` `project.dependencies` section as dependency format (per [poetry docs](https://python-poetry.org/docs/managing-dependencies/))
 - Add `poetry` test harness
 - Add `poetry` to CI
 - Update documentation

**Also bumps version in preparation for release**

Closes #375

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,6 +81,11 @@ jobs:
           cache: false
           run-install: false
   
+      - name: Set up Poetry
+        uses: abatilo/actions-poetry@v3
+        with:
+          poetry-version: latest
+
       - name: Cache conda packages
         uses: actions/cache@v4
         env:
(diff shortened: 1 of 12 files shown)


jayqi (approved):
Looks good.

I didn't realize Poetry finally added support for PEP 621 `[project]` metadata late last year, finally.

One thing to note is that we might want to swap in poetry-core as the build backend, instead of flit-core. I don't think this is strictly necessary since we don't do anything Poetry-specific in our generated `pyproject.toml`. I've confirmed that it works as is with flit-core, since the package we generate follows the common part of the specification.

pjbull:
> One thing to note is that we might want to swap in poetry-core as the build backend, i

Makes sense. I updated that. If tests pass, I will merge and release.

PR #470, opened 2025-12-05
Very ... very opinionated fork with jupyterlab goodies and support to ... well ... copier

stellarshenson:
Hi you will likely not like this very much, but my work as a data scientist and engineer over the years - and also the work of my team has resulted in this fork. Which is a very opinionated one with the template migration to copier

The Readme file and philosophy document provides you with the background behind this fork. I would be happy to have a chat about it - we can jump on a call if you can


pjbull:
Thanks @stellarshenson, cool to see other related projects inspired by CCDS. The right place for this would be to share it in "show and tell" on the discussions page so other folks can find it: https://github.com/drivendataorg/cookiecutter-data-science/discussions/categories/show-and-tell

Going to close out the PR, since we won't merge it upstream.

PR #473, opened 2025-12-29
Tweak test harnesses to address Windows actions runner failures

### LINKED ISSUE #432: Windows runners fail sporadically during conda tests
Example of a failed job [here](https://github.com/drivendataorg/cookiecutter-data-science/actions/runs/13680919847/job/38254099375).

The Windows test runners will fail occasionally when running the tests. They usually fail on a PR run when all the windows runners are run at once, and re-running each individually tends to work out okay.

It seems to always fail on one of the conda environment tests.

The specific failure is a Windows OS error:

```
make[1]: *** [Makefile:17: requirements] Error -1073741819
```

This appears to be a [Windows file system error](https://learn.microsoft.com/en-us/answers/questions/379441/error-code-exited-with-code-1073741819).

chrisjkuch:
Closes #432 

This PR makes two small tweaks to the ~Makefile~ Windows environment config in the actions test runners to address intermittent errors on Windows runners:

* Give conda on Windows runners some breathing room to settle by sleeping for 2 seconds
* Remove phantom ruff configs that supercede pyproject.toml on windows

### DIFF
--- a/tests/pixi_harness.sh
+++ b/tests/pixi_harness.sh
@@ -51,11 +51,15 @@ if [ -d ".pixi" ]; then
     echo "Removing Ruff config files under .pixi to avoid discovery"
     find .pixi \( -name ".ruff.toml" -o -name "ruff.toml" \) -print -delete
 fi
-pixi run python - <<'PY'
+RUFF_PATH="$(pixi run python - <<'PY'
 import shutil
-print(f"ruff which: {shutil.which('ruff')}")
+print(shutil.which("ruff") or "")
 PY
-pixi run ruff --version
+)"
+echo "ruff which: ${RUFF_PATH:-None}"
+if [ -n "$RUFF_PATH" ]; then
+    pixi run ruff --version
+fi
 
 pixi run make lint
 pixi run make format
--- a/tests/pixi_harness.sh
+++ b/tests/pixi_harness.sh
@@ -22,6 +22,9 @@ function finish {
     if [ -f "pixi.lock" ]; then
         rm -f pixi.lock
     fi
+    if [ -f ".ruff-empty.toml" ]; then
+        rm -f .ruff-empty.toml
+    fi
 }
 trap finish EXIT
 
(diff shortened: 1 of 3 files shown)


jayqi (changes requested):
I'm pretty suspicious about these changes, because they don't feel like we should be making them.

`--prune` is necessary for syncing the conda environment with what is specified in the `environment.yml`. Removing it changes the behavior. If we think there is a race condition, then we should try to try to synchronize or put in a delay or something instead of changing the behavior of the command. 

For the ruff configuration, if we have it set up correctly, then it should find the configuration in `pyproject.toml` automatically. If it's not working, that feels like a symptom of a different problem.

chrisjkuch:
@jayqi this is ready for another pass

chrisjkuch:
[nu] @jayqi when you get a chance can you re-review? The windows tests in other PRs are still failing intermittently, this appears to make it better though may not solve it completely.

tests/pixi_harness.sh
```diff
@@ -46,7 +46,21 @@ if [ -f "$MODULE_NAME/config.py" ]; then
     pixi run python -c "from $MODULE_NAME import config"
 fi
 
-# Run linting and formatting through pixi
+# Run linting/formatting through make, but remove any .pixi ruff configs that Ruff may discover.
```
jayqi, line 49: @chrisjkuch what is going on here?

chrisjkuch: Okay this one is really odd. I think what happens is that there's a .ruff.toml file somewhere that gets picked up as the default ruff formatter instead of the project.toml file, and no matter what I did I couldn't figure out a way to get rid of it other than to just remove it when it was getting installed. I have some LLM convos I can try to dig up around this.

chrisjkuch: Here's a link to a failing job demonstrating what this is trying to fix: https://github.com/drivendataorg/cookiecutter-data-science/actions/runs/22636973625/job/65602434162#step:11:1

This job is for the AGENTS.md file change and does not include the code in this PR.

Note the .ruff.toml it finds and latches onto but can't seem to read? This is trying to address whatever failure mode that is.

jayqi: You're looking at an error like this?

```
ruff failed
  Cause: Failed to load extended configuration `C:\\Users\\runneradmin\\AppData\\Local\\Temp\\tmpnpzbjkl3data-project\\my-test-repo-16\\.pixi\\envs\\default\\.ruff.toml` (`C:\\Users\\runneradmin\\AppData\\Local\\Temp\\tmpnpzbjkl3data-project\\my-test-repo-16\\.pixi\\envs\\default\\Tools\\i18n\\.ruff.toml` extends `C:\\Users\\runneradmin\\AppData\\Local\\Temp\\tmpnpzbjkl3data-project\\my-test-repo-16\\.pixi\\envs\\default\\.ruff.toml`)
  Cause: Failed to read C:\\Users\\runneradmin\\AppData\\Local\\Temp\\tmpnpzbjkl3data-project\\my-test-repo-16\\.pixi\\envs\\default\\.ruff.toml
  Cause: The system cannot find the file specified. (os error 2)
```

All of these messages just point to the same file. `pixi\envs\default` is the path to the default conda environment that Pixi creates. The way I read this error message is that it's trying to find a `.ruff.toml` there (for some reason) as if it's the current working directory,  or as if it was told to go look for a configuration in that directory, and then just doesn't find it. 

It's definitely weird.

I wonder if this is because we do `pixi run make lint` and make does something weird.

chrisjkuch: Yeah, it's only doing this on the windows runners and it's intermittent. When that file does exist, it gets used by the formatter and fails - when it doesn't, it crashes like this.

PR #476, opened 2026-03-03
Add AGENTS.md template

chrisjkuch:
Closes #474 

This PR adds an AGENTS.md file that concisely encodes desired agent interaction with the codebase. We incorporate the opinions from the opinions page and dynamically add sections based on the user's selected tooling.

### DIFF
--- a/{{ cookiecutter.repo_name }}/AGENTS.md
+++ b/{{ cookiecutter.repo_name }}/AGENTS.md
@@ -52,9 +52,9 @@ The `data/` and `models/` directories are gitignored. Do not commit data files, 
 
 Run tasks through `make`. Available recipes:
 
-* `make` — List all available commands.
-* `make requirements` — Install/update dependencies.
-* `make data` — Run the data processing pipeline.{% if cookiecutter.linting_and_formatting != 'none' %}
+* `make` — List all available commands
+* `make requirements` — Install/update dependencies
+* `make data` — Run the data processing pipeline (assumes data is present in `data/raw/`){% if cookiecutter.linting_and_formatting != 'none' %}
 * `make lint` — Check code style.
 * `make format` — Auto-format code.{% endif %}{% if cookiecutter.testing_framework != 'none' %}
 * `make test` — Run the test suite.{% endif %}{% if not cookiecutter.dataset_storage.none %}
@@ -67,7 +67,7 @@ Add project-specific recipes to the `Makefile` for commands that are run frequen
 
 ### Notebooks are for exploration; source code is for repetition
 
-Use `notebooks/` for exploratory analysis. When code is reused across notebooks, refactor it into the `{{ cookiecutter.module_name }}/` package. The project is installed as a local package, so you can import with:
+Use `notebooks/` for exploratory analysis notebooks. When code is reused across notebooks, refactor it into the `{{ cookiecutter.module_name }}/` package. The project is installed as a local package, so you can import with:
 
 ```python
 from {{ cookiecutter.module_name }}.dataset import main
@@ -77,7 +77,7 @@ Notebook naming convention: `<step>.<order>-<identifier>-<description>.ipynb` (e
 
 ### Secrets
 
-**NEVER** read `.env` or any secrets files directly.. Code should load secrets with `python-dotenv`. Use `{{ cookiecutter.module_name }}/config.py` for project paths and configuration. Never hardcode credentials, print secrets in logs, or add them to source control.
+**NEVER** read `.env` or any secrets files directly. Code should load secrets with `python-dotenv`. Use `{{ cookiecutter.module_name }}/config.py` for project paths and configuration. Never hardcode credentials, print secrets in logs, or add them to source control.
 
 ## Development Workflow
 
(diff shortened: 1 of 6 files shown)


{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+```python
+from {{ cookiecutter.module_name }}.dataset import main
+```
+
+Notebook naming convention: `<step>-<identifier>-<description>.ipynb` (e.g., `0.3-bull-visualize-distributions.ipynb`). Step numbers: 0=exploration, 1=cleaning/features, 2=visualizations, 3=modeling, 4=publication.
+
+### Secrets
+
+Store secrets in `.env` (gitignored), and do not ever read the `.env` file directly. Load secrets with `python-dotenv`. Use `{{ cookiecutter.module_name }}/config.py` for project paths and configuration. Never hardcode credentials or put them in source control.
```
jayqi: I think the agent generally shouldn't be handling secrets, so the first sentence should be more descriptive rather than imperative.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
+# AGENTS.md
+
+**Project name:** {{ cookiecutter.project_name }}
+**Description:** {{ cookiecutter.description }}
```
jayqi: You have a lot of single newline delimited lists, but markdown does not treat those as actual line breaks, which makes it less human-readable. I think it makes more sense to do unordered lists here.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+Treat the data pipeline as a directed acyclic graph. Each step takes inputs and produces outputs with no circular dependencies. Anyone must be able to reproduce final outputs from code and raw data alone.
+
+### Raw data is immutable
+
+Never edit, overwrite, or manually modify files in `data/raw/`. Data flows one direction:
+
+`data/raw/` → `data/interim/` → `data/processed/`
+
+Intermediate outputs should be cached in `interim/`. Final products go to `processed/`.
```
pjbull: > "Final products"

I think of this more as "analysis-ready datasets" that don't need anymore preprocessing or feature engineering.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+│   ├── config.py     ← Project configuration and path definitions.
+│   ├── dataset.py    ← Data loading and generation.
+│   ├── features.py   ← Feature engineering code.
+│   ├── plots.py      ← Visualization code.
+│   └── modeling/
+│       ├── train.py  ← Model training.
+│       └── predict.py← Model inference.
+{%- endif %}
+└── tests/            ← Test suite.
+```
```
jayqi: Do LLMs understand visual file trees like this reliably? Maybe worth testing.

jayqi (changes requested):

{{ cookiecutter.repo_name }}/AGENTS.md
pjbull: or print them in logs

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+{% if cookiecutter.linting_and_formatting == 'ruff' %}
+**Linting/Formatting:** Uses ruff. Run `make lint` to check, `make format` to fix.
+{%- elif cookiecutter.linting_and_formatting == 'flake8+black+isort' %}
+**Linting/Formatting:** Uses flake8, black, and isort. Run `make lint` to check, `make format` to fix.
+{%- endif %}
+{% if cookiecutter.testing_framework == 'pytest' %}
+**Testing:** Uses pytest. Run `make test`.
+{%- elif cookiecutter.testing_framework == 'unittest' %}
+**Testing:** Uses unittest. Run `make test`.
+{%- endif %}
```
pjbull, line 122: May want instructions to ensure that linting and tests pass before committing.

pjbull (commented):
Great place to start! A few small comments.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+Never edit, overwrite, or manually modify files in `data/raw/`. Data flows one direction:
+
+`data/raw/` → `data/interim/` → `data/processed/`
+
+Intermediate outputs should be cached in `interim/`. Final products go to `processed/`.
+
+### Data is not in source control
+
+The `data/` and `models/` directories are gitignored. Do not commit data files, trained models, or `.env` files to git.{% if not cookiecutter.dataset_storage.none %} Use `make sync_data_down` / `make sync_data_up` to sync data with cloud storage. Prefer using `cloudpathlib` in code to interact with cloud storage in a consistent way.{% endif %}
```
pjbull: > Prefer using `cloudpathlib` in code to interact with cloud storage in a consistent way.

I think that the best practice is either one of the following:

 - Something like `S3Client(local_cache_dir="./data").set_as_default_client()` in the settings/config and always use cloudpathlib. Let cloudpathlib handle the caching of files to that dir.
 - OR, use the sync_data_up and sync_data_down commands and treat data as local.

It's an antipattern to do both because you lose the thread on what is canonical.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Notebooks are for exploration; source code is for repetition
+
+Use `notebooks/` for exploratory analysis. When code is reused across notebooks, refactor it into the `{{ cookiecutter.module_name }}/` package. The project is installed as a local package, so you can import with:
+
+```python
+from {{ cookiecutter.module_name }}.dataset import main
+```
+
+Notebook naming convention: `<step>-<identifier>-<description>.ipynb` (e.g., `0.3-bull-visualize-distributions.ipynb`). Step numbers: 0=exploration, 1=cleaning/features, 2=visualizations, 3=modeling, 4=publication.
```
jdcc: This doesn't describe the point numbers for each step.

{{ cookiecutter.repo_name }}/AGENTS.md
```diff
@@ -0,0 +1,121 @@
...
+- `make data` — Run the data processing pipeline.{% if cookiecutter.linting_and_formatting != 'none' %}
+- `make lint` — Check code style.
+- `make format` — Auto-format code.{% endif %}{% if cookiecutter.testing_framework != 'none' %}
+- `make test` — Run the test suite.{% endif %}{% if not cookiecutter.dataset_storage.none %}
+- `make sync_data_down` — Pull data from cloud storage.
+- `make sync_data_up` — Push data to cloud storage.{% endif %}
+- `make create_environment` — Set up the Python environment.
+- `make clean` — Remove compiled Python files.
+
+Add project-specific recipes to the `Makefile` as needed.
```
jdcc: Can we be any more specific with examples here? Model training, eval, figures? My concern is that this is either ignored or over-adhered to.

jdcc (commented):
A couple small things, but I don't feel like I know enough to weigh in otherwise. Do we want to flag to folks setting up a project that they can/should fill in project-specific context in here?

{{ cookiecutter.repo_name }}/AGENTS.md
chrisjkuch: Given that the rest of the template currently uses the `make sync` commands and cloudpathlib isn't in the requirements for basic python packages, I'm planning to remove the second sentence. If we feel strongly enough that only cloudpathlib should be used to interact with cloud files, I think we can file an issue / handle in a separate PR.

chrisjkuch:
@jdcc @pjbull @jayqi I've responded to feedback, would love you to take another look when you have a chance!

